PR #1, opened 2018-01-27
copy paste my code over to the new provider layout

HX-Rd:
This is basically just a copy paste so you might have to look it over


adamdecaf:
Awesome, thanks!

PR #3, opened 2018-06-08
Add NS provider

lukegiuliani:
Hiya,

Basically i've just pulled the code from https://github.com/matthewbentley/terraform-provider-namecheap, which was in turn taken from https://github.com/HX-Rd/terraform/tree/f-namecheap-provider, and pushed it into your repo. 

I'm not sure what has to happen to the reference in main.go when it gets back to your repo, hence leaving in the comment. 

Would love it if you accept and do a release so I can update my docs to the newly NS supporting provider :)

Thanks,

LG

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,8 +1,9 @@
 package main
 
 import (
-	"github.com/adamdecaf/terraform-provider-namecheap/namecheap"
+	//	"github.com/adamdecaf/terraform-provider-namecheap/namecheap"
 	"github.com/hashicorp/terraform/plugin"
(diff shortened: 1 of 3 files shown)


adamdecaf:
@lukegiuliani I released `1.1.0-dev`, can you try it out? https://github.com/adamdecaf/terraform-provider-namecheap/releases/tag/1.1.0-dev

I'm having a bit of a problem using it locally. I can't get the setup for non-official providers setup. https://www.terraform.io/docs/configuration/providers.html#third-party-plugins 

If you can verify it works great! Let me know what you had to do.

PR #10, opened 2018-12-09
Added the constants you have to the actual defaults. Added timeout to API.

OGProgrammer:
* Added the consts for ttl, mx pref, and timeout default to the schema default value.
* Added the timeout feature on Terraform to 30 seconds for API calls to NC. 
* Added a build and install script but debating on removing if owner wants to from PR.

### DIFF
--- a/namecheap/resource_namecheap_record.go
+++ b/namecheap/resource_namecheap_record.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strconv"
 	"sync"
+	"time"
 
 	"github.com/adamdecaf/namecheap"
 	"github.com/hashicorp/terraform/helper/schema"
@@ -12,8 +13,10 @@ import (
 // We need a mutex here because of the underlying api
 var mutex = &sync.Mutex{}
 
-const ncDefaultTTL int = 1800
+// This is the "Auto" TTL setting in Namecheap
+const ncDefaultTTL int = 1799
 const ncDefaultMXPref int = 10
+const ncDefaultTimeout time.Duration = 30
 
 func resourceNameCheapRecord() *schema.Resource {
 	return &schema.Resource{
@@ -21,6 +24,13 @@ func resourceNameCheapRecord() *schema.Resource {
 		Update: resourceNameCheapRecordUpdate,
 		Read:   resourceNameCheapRecordRead,
 		Delete: resourceNameCheapRecordDelete,
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(ncDefaultTimeout * time.Second),
+			Update: schema.DefaultTimeout(ncDefaultTimeout * time.Second),
+			Read:   schema.DefaultTimeout(ncDefaultTimeout * time.Second),
+			Delete: schema.DefaultTimeout(ncDefaultTimeout * time.Second),
+		},
 
 		Schema: map[string]*schema.Schema{
 			"domain": &schema.Schema{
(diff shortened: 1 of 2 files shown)


namecheap/resource_namecheap_record.go
```diff
@@ -22,6 +25,13 @@ func resourceNameCheapRecord() *schema.Resource {
...
 
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(ncDefaultTimeout * time.Second),
+			Update: schema.DefaultTimeout(ncDefaultTimeout * time.Second),
+			Read:   schema.DefaultTimeout(ncDefaultTimeout * time.Second),
+			Delete: schema.DefaultTimeout(ncDefaultTimeout * time.Second),
+		},
+
 		Schema: map[string]*schema.Schema{
 			"domain": &schema.Schema{
```
OGProgrammer: BTW My IDE says these are redundant, maybe we can remove these to top level nodes?

scripts/build-install.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+make build
+mv $GOPATH/bin/terraform-provider-namecheap ~/.terraform.d/plugins/terraform-provider-namecheap
+chmod +x ~/.terraform.d/plugins/terraform-provider-namecheap
```
OGProgrammer: I'll remove this if you want. I run this from root.

namecheap/resource_namecheap_record.go
```diff
@@ -12,8 +13,10 @@ import (
 // We need a mutex here because of the underlying api
 var mutex = &sync.Mutex{}
 
-const ncDefaultTTL int = 1800
+// This is the "Auto" TTL setting in Namecheap
+const ncDefaultTTL int = 1799
```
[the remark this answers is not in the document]
adamdecaf: Weird to be off by one, but thanks!

namecheap/resource_namecheap_record.go
adamdecaf: Maybe? I probably copied the code structure from previous Terraform providers. Feel free to condense the code if you want.

scripts/build-install.sh
adamdecaf: I've got a symlink for the same reason.

```
$ ls -l ~/.terraform.d/plugins/
total 0
lrwxr-xr-x  1 adam  staff  49 Dec  4 11:58 terraform-provider-namecheap -> /Users/adam/code/bin/terraform-provider-namecheap
```

namecheap/resource_namecheap_record.go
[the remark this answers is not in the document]
OGProgrammer: Yea when you set "Auto" on their front end, it must just set it to this 1799 value really but I bet you the ttl actually varies still when on this setting. I'm not sure what namecheap means by "Auto" for TTL but maybe based on record types, it returns diff values.

scripts/build-install.sh
OGProgrammer: Should I go with the symlink or is this script ok? This really should be in the makefile..

namecheap/resource_namecheap_record.go
```diff
@@ -22,34 +25,41 @@ func resourceNameCheapRecord() *schema.Resource {
...
 				Type:     schema.TypeString,
 				Required: true,
 			},
-			"type": &schema.Schema{
+			"type": {
 				Type:     schema.TypeString,
 				Required: true,
 			},
-			"address": &schema.Schema{
+			"address": {
```
OGProgrammer: Updated, let me know if you need other changes.

namecheap/resource_namecheap_record.go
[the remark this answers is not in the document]
adamdecaf: Yea, I'm not sure. Let's stick with the Namecheap default. I lower my records to 30/60s, but the providers default should be ours IMO.

scripts/build-install.sh
adamdecaf: I'd say symlink unless a file already exists at `~/.terraform.d/plugins/terraform-provider-namecheap`

adamdecaf:
@OGProgrammer If you want I can delete that file from the PR and merge this.

OGProgrammer:
Holidays kicked my ass sorry @adamdecaf , just pushed the change so we should be good to go on merging this. Still gotta work on the throttling bug, that is effecting me with a ton of domains managed by terraform now haha.

PR #28, opened 2020-01-09
Update Mac installation instructions

stchris:
Following https://www.terraform.io/docs/configuration/providers.html#third-party-plugins it seems necessary to hint at the provider version in the binary name.


adamdecaf:
Good catch, thanks!

PR #30, opened 2020-02-04
Troubleshooting for readme (bad hash error)

OGProgrammer:
Not sure if this should be the right place for this but figured I document it. Just added a troubleshooting section for the readme with a few things I ran into today getting this going again. If we should move it to a troubleshooting.md or w/e just lmk

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ $ wget -O ~/.terraform.d/plugins/terraform-provider-namecheap https://github.com
 ```bash
 $ mkdir -p ~/.terraform.d/plugins/
 $ curl -L https://github.com/adamdecaf/terraform-provider-namecheap/releases/download/1.5.0/terraform-provider-namecheap-osx-amd64 > ~/.terraform.d/plugins/terraform-provider-namecheap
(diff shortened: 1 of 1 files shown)


adamdecaf:
We could refer them to `terraform state rm`

```
$ terraform state rm --help 
Usage: terraform state rm [options] ADDRESS...

  Remove one or more items from the Terraform state, causing Terraform to
  "forget" those items without first destroying them in the remote system.

  This command removes one or more resource instances from the Terraform state
  based on the addresses given. You can view and list the available instances
  with "terraform state list".

  If you give the address of an entire module then all of the instances in
  that module and any of its child modules will be removed from the state.

  If you give the address of a resource that has "count" or "for_each" set,
  all of the instances of that resource will be removed from the state.

Options:

  -dry-run            If set, prints out what would've been removed but
                      doesn't actually remove anything.

  -backup=PATH        Path where Terraform should write the backup
                      state.

  -lock=true          Lock the state file when locking is supported.

  -lock-timeout=0s    Duration to retry a state lock.

  -state=PATH         Path to the state file to update. Defaults to the current
                      workspace state.

```

OGProgrammer:
Updated

PR #33, opened 2020-07-15
Update README.md

b-m-f:
Had to add the version on Linux to get Terraform to pick up the provider. Maybe this can be automatically set to the latest release somehow?

Linux: 5.6.16-1-MANJARO


adamdecaf:
Thanks! It should be automated, but there aren't a lot of releases with this project.

PR #36, opened 2020-12-31
Added Windows install instructions to README.md

Jonas-Sander:
This will add the `terraform-provider-namecheap.exe` to `%APPDATA%\terraform.d\plugins\registry.terraform.io\hashicorp\namecheap\1.5.0\windows_amd64\terraform-provider-namecheap_v1.5.0.exe`. 

In `-OutFile ( New-Item -Path $env:APPDATA\terraform.d\plugins\registry.terraform.io\hashicorp\namecheap\1.5.0\windows_amd64\terraform-provider-namecheap_v1.5.0.exe -Force )` the `-Force` flag will create the directory if it does not exist.

If there is any easier way then let me know it took me like 2 hours to figure this out because the documentation is so sparse.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,11 @@ $ curl -L https://github.com/adamdecaf/terraform-provider-namecheap/releases/dow
 $ chmod +x ~/.terraform.d/plugins/terraform-provider-namecheap
 ```
 
+## Windows
+```powershell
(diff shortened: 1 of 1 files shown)


ifree92:
Thanks for contributing.
We have made complete project refactoring #46 

Soon we will provide detailed development instructions for all existing operating systems.

Issue #46, opened 2021-06-29
[ATTENTION] Complete project refactoring


ifree92:
Hi there,

We want to update our community about the further development of the project.

Short history: about a month before we have transferred the project under our support at @namecheap from @adamdecaf.
Really thanks to the author and all contributors for your time and efforts, much appreciated. Now we will do our best to get it stable asap :)

Upon our learning and debugging of the existing codebase, we have faced a few critical issues that require the following complete refactoring. Before going through we want to receive some feedback from our community.

### Current approach & issues

Example:

```tf
resource "namecheap_record" "www-example-com" {
  name = "www"
  domain = "example.com"
  address = "127.0.0.1"
  type = "A"
}

resource "namecheap_record" "www-example2-com" {
  name = "www"
  domain = "example2.com"
  address = "127.0.0.1"
  type = "A"
}

resource "namecheap_ns" "www-mydomain-com" {
  domain = "mydomain.com"
  servers = ["ns1.domain.com", "ns2.domain.com"]
}
```

Currently, we have the approach where each resource item reflects a record for a specific domain (including nameserver record). [Our API "setHosts"](https://www.namecheap.com/support/api/methods/domains-dns/set-hosts/) requires to send back all existing records with(out) amended/created/removed one. This compels to use a mutex to prevent racing, decreases general performance, and probably might be a part of bug #31. One more issue #37 related to requirements of ["setHosts" API](https://www.namecheap.com/support/api/methods/domains-dns/set-hosts/) to provide the `EmailType` mode with possible values NONE/FWD/MXE/MX/OX that will allow to set up email behavior. In case we will create MX record, we might face an infinity loop (probably one more reason for #31 when we have added an MX record, we're sending a request to create, then upon reading record still doesn't exist and here we have the problem).

### Refactoring proposal

Example:

```tf
resource "namecheap_domain_records" "www-example-com" {
  domain = "example.com"
  email_type = "MX"

  record {
    hostname = "host1"
    type = "A"
    address = "10.11.12.13"
    ttl = 1800
  }

  record {
    hostname = "host2"
    type = "A"
    address = "11.12.13.14"
  }

  record {
    hostname = "mail"
    type = "MX"
    address = "mail.domain.com"
    mx_pref = 10
    ttl = 1800
  }
}

resource "namecheap_domain_records" "www-example2-com" {
  domain = "example2.com"
  nameservers = ["ns1.domain.com", "ns2.domain.com"]
}
```

[...]

Issue #37, opened 2021-01-06
Unable to create MX records


leveled:
I seem to be unable to create MX records. Below is a redacted, abbreviated version of my terraform files. I've tried both with appended '.' and no '.' and it does not seem to make a difference. Some of these variables are not relevant to MX records, would omitting them help? I'd like to use the same generic module for record creation but can make a separate one if that is required. 

```
terraform {
[...]

#Add MX Record
module "namecheap_MX_record" {
  source = "./tf_modules/namecheap/create-record"
  type = "MX"
  name = "mail.${var.domain}."
  domain = var.domain
  address = "mail.${var.domain}."
  mx_pref = "10"
  ttl = "3600"

[...]

#Specifies namecheap credentials
provider "namecheap" {
  username = var.namecheap_username
  api_user = var.namecheap_api_user
  token = var.namecheap_api_token
  ip = var.namecheap_ip_address
}

#Create record
resource "namecheap_record" "record" {
  name = var.name
  domain = var.domain
  type = var.type
  address = var.address
  mx_pref = var.mx_pref
  ttl = var.ttl
}


}
```

ifree92:
Thanks for reporting.
The issue has been completely resolved in the new v2 major release.

Please have a look https://registry.terraform.io/providers/namecheap/namecheap/latest/docs/guides/namecheap_domain_records_guide#email-type to get details about MX records.

Issue #32, opened 2020-05-14
Failed to create namecheap Record: Invalid record type


cyberbutler:
### Terraform Version
```
Terraform v0.12.25
+ provider.http v1.2.0
+ provider.namecheap (unversioned)
```
### Namecheap provider version

Both built from source and Release v1.5.0

### Affected Resource(s)
- namecheap_record

### Terraform Configuration Files
```hcl
variable nc_username {
	type = string
	default = ""
}
variable nc_token {
	type = string
	default = ""
}
data http "myip" {
	url = "https://ipinfo.io/ip"
}

provider namecheap {
	use_sandbox = false
	username = var.nc_username
	api_user = var.nc_username
	token = var.nc_token
	ip = data.http.myip.body
}

resource namecheap_record "record" {
	type = "A"
	name = "test"
	domain = "example.com"
        address = "127.0.0.1"
}
```

### Debug Output
https://gist.github.com/cyberbutler/39753094328b7a623a5ebc5004e507ac

### Expected Behavior
What should have happened?
Valid Record Type `A` should have been passed to the Namecheap Go API Successfully

### Actual Behavior
What actually happened?
```
Error: Failed to create namecheap Record: Invalid record type, allowed types="A, AAAA, ALIAS, CAA, CNAME, MX, MXE, TXT, URL, URL301, FRAME"

  on namecheap.tf line 1, in resource "namecheap_record" "record":
   1: resource namecheap_record "record" {
```
### Steps to Reproduce
1. Configure the above terraform resources
2. `terraform apply`

### Important Factoids
Attempted to run from MacOS and Debian systems, both results in the same error.
I've also run with both the released provider and built the provider from the `master` branch here. Same results. 

I tracked down the provided error [here](https://github.com/adamdecaf/namecheap/blob/d47b19511bf1d39c7482d2ac3420016de7050b73/host.go#L40). I also added some janky `fmt.Printf` "debug" code into the local build of the provider to see what was held in [`RecordType` before it was passed to the `namecheap.client`](https://github.com/adamdecaf/terraform-provider-namecheap/blob/master/namecheap/resource_namecheap_record.go#L82) but it didn't seem to be wrong in anyway. Perhaps there is a user error here somewhere?

cyberbutler:
Just to be more concise the RecordType does not pass the `CheckRecordType` validation in the `namecheap` module. I cannot figure out why, I've also replicated the validation code with no issues:

```go
package main

import (
	"fmt"
[...]

Issue #31, opened 2020-03-04
root is waiting for provider namecheap


leylmordor:
Been trying to terraform apply it but i keep getting an error. 
Any ideas ?

Versions:
**Terraform v0.12.20**
**Namcheap v1.4.0**

Here is the error
```
2020/03/04 09:56:34 [TRACE] dag/walk: vertex "root" is waiting for "provider.namecheap (close)"
2020/03/04 09:56:34 [TRACE] dag/walk: vertex "meta.count-boundary (EachMode fixup)" is waiting for "module.domainOne-com.module.domainOne-com-ns-two.namecheap_record.record"
2020/03/04 09:56:34 [TRACE] dag/walk: vertex "provider.namecheap (close)" is waiting for "module.domainTwo-com.module.domainTwo-com-a-two.namecheap_record.record"
2020/03/04 09:56:39 [TRACE] dag/walk: vertex "meta.count-boundary (EachMode fixup)" is waiting for "module.domainOne-com.module.domainOne-com-ns-two.namecheap_record.record"
2020/03/04 09:56:39 [TRACE] dag/walk: vertex "provider.namecheap (close)" is waiting for "module.domainTwo-com.module.domainTwo-com-a-two.namecheap_record.record"
2020/03/04 09:56:39 [TRACE] dag/walk: vertex "root" is waiting for "provider.namecheap (close)"
module.domainTwo-com.module.domainTwo-com-txt-one.namecheap_record.record: Still creating... [30s elapsed]
module.domainOne-com.module.domainOne-com-ns-one.namecheap_record.record: Still creating... [30s elapsed]
module.domainTwo-com.module.domainTwo-com-cname-ne.namecheap_record.record: Still creating... [30s elapsed]
module.domainTwo-com.module.domainTwo-com-a-three.namecheap_record.record: Still creating... [30s elapsed]
module.domainOne-com.module.domainOne-com-ns-four.namecheap_record.record: Still creating... [30s elapsed]
module.domainThree-com.module.domainThree-com-url-one.namecheap_record.record: Still creating... [30s elapsed]
module.domainTwo-com.module.domainTwo-com-a-one.namecheap_record.record: Still creating... [30s elapsed]
module.domainTwo-com.module.domainTwo-com-mx-two.namecheap_record.record: Still creating... [30s elapsed]
module.domainTwo-com.module.domainTwo-com-mx-one.namecheap_record.record: Still creating... [30s elapsed]
[...]

PR #47, opened 2021-07-07
Refactor/complete project refactoring

ifree92:
This PR reflects all nuances described in #46.

This is a completely new terraform provider with breaking changes.

Features:
- MERGE and OVERWRITE mode as discussed on #46.
- parallel resources processing (got rid of mutexes) that is the fix of #31 
- better input data validation that is the fix of #32
- ability to change `email_mode` (refer to https://www.namecheap.com/support/api/methods/domains-dns/set-hosts/ `EmailType` request parameter) that is the fix of #37 
- stability improvements

Remaining work:
- on create, if default parking records are set, then remove them before applying the resource ✅
- tests (next PR)
- documentation & examples (next PR)


namecheap/provider.go
```diff
@@ -1,109 +1,66 @@
...
-				DefaultFunc: schema.EnvDefaultFunc("NAMECHEAP_API_USER", nil),
-				Description: "A registered apiuser for namecheap",
+				Description: "A registered api user for namecheap",
 			},
 
-			"token": {
+			"api_key": {
 				Type:        schema.TypeString,
 				Required:    true,
-				DefaultFunc: schema.EnvDefaultFunc("NAMECHEAP_TOKEN", nil),
```
StyleT, line 45: I think we should support passing credentials as env variables

StyleT (changes requested):

namecheap/provider.go
ifree92: Got it.
Reverted back

namecheap/namecheap_domain_record.go
```diff
@@ -0,0 +1,375 @@
...
+
+	if recordsRaw, ok := data.GetOk("record"); ok {
+		records = recordsRaw.(*schema.Set).List()
+	}
+
+	if nameserversRaw, ok := data.GetOk("nameservers"); ok {
+		nameservers = nameserversRaw.([]interface{})
+	}
+
+	if mode == "MERGE" && records != nil {
```
adamdecaf, line 118: Do we want to use `strings.EqualFold` instead of case sensitive equality?

adamdecaf:
This is looking great!

namecheap/namecheap_domain_record.go
ifree92: I thought about it to set it as non case sensitive.
The only reason is consistency, since other fields (e.g. record.type field or email_type) requires values in upper case, so that's why I leaved it in upper case.

However, from developer experience perspective, probably you're right.
If you think it is better, I will easily amend it

ifree92: @adamdecaf I have made a few changes. Now mode & records comparison is safer.
Thanks for the comments! :)

StyleT (approved):

adamdecaf (approved):

PR #48, opened 2021-07-13
feat: added email type GMAIL support

ifree92:


namecheap/namecheap_domain_record.go
```diff
@@ -14,11 +14,12 @@ const (
 	ncModeMerge     = "MERGE"
 	ncModeOverwrite = "OVERWRITE"
 
-	ncEmailTypeNONE = "NONE"
-	ncEmailTypeFWD  = "FWD"
-	ncEmailTypeMXE  = "MXE"
-	ncEmailTypeMX   = "MX"
-	ncEmailTypeOX   = "OX"
+	ncEmailTypeNONE  = "NONE"
```
StyleT, line 17: Gould we fetch this list from sdk? How we have hardcoded values in both repos

StyleT (changes requested):

PR #52, opened 2021-07-16
Fix/smart records merge on create

ifree92:


docs/resources/domain_records.md
```diff
@@ -60,3 +60,5 @@ resource "namecheap_domain_records" "my-domain2-com" {
 - `type` - (Required) Possible values: A, AAAA, ALIAS, CAA, CNAME, MX, MXE, NS, TXT, URL, URL301, FRAME
 - `mx_pref` - (Optional) MX preference for host. Applicable for MX records only
 - `ttl` - (Optional) Time to live for all record types. Possible values: any value between 60 to 60000
+
+~> It is strongly recommended to set `address` and `hostname` in lower case to prevent undefined behavior!
```
StyleT, line 64: Can we always convert them to lowercase automatically during comparison?

docs/guides/namecheap_domain_records_guide.md
```diff
@@ -34,6 +34,9 @@ The same workflow works for both: `record` items and `nameservers`.
 
 -> This is the default behavior, however, we recommend to set the mode explicitly.
 
+~> Upon creating a new resource with the records or nameservers that already exist, the provider will throw a duplicate
```
StyleT, line 37: I think we need to add that it happens if subdomain & record type match

StyleT (changes requested):
Let's cover existing code with some basic tests & merge this PR with tests that are covering the issue.

docs/guides/namecheap_domain_records_guide.md
ifree92: Noted. I will check this.
At the moment, we have duplicates comparison only by `hostname` (subdomain), `type`, and `address`.

However, user allowed to create "mistakes" like:

```terraform
resource "namecheap_domain_records" "domain" {
   #...
   record {
     hostname = "dev"
     type = "A"
     address = "11.22.33.44"
   }

   record {
     hostname = "dev"
     type = "A"
     address = "33.33.33.33"
   }
}
```

I will amend this behavior (and provide tests for all cases! 😆 )

docs/resources/domain_records.md
ifree92: Basically, yes.
However, the existing behavior works in the way where if a user changes the camel case of the address of any record, the provider will "re-create" the record. I will have to make the READ smarter.

I will revert on this in the context of a current PR.

.github/workflows/ci.yml
```diff
@@ -24,6 +25,100 @@ jobs:
...
+          go version
+          go env GOPATH
+
+      - name: Acceptance Test
+        run: |
+          export CGO_ENABLED=0
+          source go-env.sh
+          make testacc
+        env:
+          NAMECHEAP_USER_NAME: ${{ secrets.NAMECHEAP_USER_NAME }}
```
StyleT, line 98: Let's move "non-secret" var values to this file. It will simplify debugging

StyleT (approved):

Issue #63, opened 2021-12-15
Provider 2.0.0+ Import?


ptous:
Hi, there. As this is a "pure" usage / documentation question and not about a failure (yet!), I'm going to start with only a little of the usual requested debug log info. If you need me to produce any debug output to answer my question though, I will gladly do so.

In short: is it possible to use the modern provider to perform a `terraform import` of an existing domain configuration? If it _is_ possible, what is the correct input for the domain "id" to provide the `terraform import` command?

I _did_ check the documentation, but there is no reference to an import command, which seems unusual. I am not sure whether I should infer from this that `import` is not viable. If not, the documentation should probably say this explicitly.

Background: 
```
$ terraform -version
Terraform v1.1.0
on linux_amd64
+ provider registry.terraform.io/namecheap/namecheap v2.0.2
```
- I am attempting to use the namecheap provider to import state for an already configured domain.
- I am **not** attempting to use the import process to migrate a pre-2.0.0 Terraform configuration - this is a new config started with the 2.0.2 provider.
- I have working API access and have confirmed that my TF config can _modify_ domains I own (tested with `mode = "MERGE"`). I also performed a negative test and confirmed that if I intentionally mangle my credentials in my provider config I receive sensible error messages to that effect.

Example `import` commands I have tried:
- `terraform import -no-color namecheap_domain_records.some_domain example.com`
- `terraform import -no-color namecheap_domain_records.some_domain 1234567`

In all cases, I receive this message:
```
Error: invalid domain: incorrect format
```

In the examples above I of course supplied real domain names and ID numbers. The numeric value used in the 2nd example was the internal "ID" field of the domain, as returned via `curl` against the API. I am fairly sure using the numeric ID was wrong, since the `id` field's value ends up the same as the `domain` field's, as viewed in `terraform state show` after using `terraform apply` on another domain. Given that, though, the response that the string domain name is in the "incorrect format" is baffling.

Thanks and best regards!

trilom:
Looked at this today and found the following:  

This line is where the issue happens:  
https://github.com/namecheap/terraform-provider-namecheap/blob/46977b868c8f9386f0c1a82b9f7465339a616286/namecheap/namecheap_domain_record.go#L193-L196

It calls the `GetList` endpoint which calls `ParseDomain`
[...]

PR #64, opened 2021-12-27
handle imports a little better

trilom:
Issue #63 talks about this, this solved it for me locally and allowed me to import with the domain name


StyleT:
Released as v2.1.0 thank you for your contribution 💪

Issue #73, opened 2023-04-30
Error if CAA 'iodef' record defined


bgre033:
Hi,

It seems if a zone in Namecheap contains a CAA 'iodef' record, the Namecheap Terraform provider fails to create DNS records. If the CAA record is removed, the creation succeeds. I've tested with A and CNAME records, using Terraform 1.4.0 and 1.4.6, with Namecheap provider 2.1.0.

### Terraform Version
Terraform v1.4.6
+ provider registry.terraform.io/namecheap/namecheap v2.1.0

### Namecheap provider version
What version of terraform-provider-namecheap are you using? 2.1.0

### Steps to Reproduce

1. Create a CAA record such as below in the zone file on Namecheap

```
CAA Record
@
iodef
"mailto:support@domain.com"
```

2. Create Namecheap resource block

```
resource "namecheap_domain_records" "create-record" {

    domain = "domain.com"
    mode = "MERGE"

    # ACM Certificate Validation Record
    record {
        hostname = "test"
        type = "A"
        address = "1.1.1.1"
        ttl = 1800
    }
}
```

3. Run 'terraform apply'
4. Output is an error such as below.

> **Error**: Records[10].Address "0 iodef "mailto:support@domain.com"" **must contain a protocol prefix for CAA iodef record**

psa:
I'm not sure why they're not merging it, but https://github.com/namecheap/terraform-provider-namecheap/pull/66 has the fix you're looking for.

bgre033:
Thanks, and good work! I have a case open with Namecheap and have passed this on. Fingers crossed they actually do something about it.

LarsArtmann:
**I just tried to add the following records and failed with the error message below.
If I add it in the Web GUI it works without any complaints.**

**I'm sure somehow the validation logic does not recognise `mailto:` correctly.**

## Code
```hcl
 record {
    address  = "0 issue \"pki.goog\""
    hostname = "@"
    mx_pref  = 10
    ttl      = 1799
    type     = "CAA"
  }
  record {
    address  = "0 iodef \"mailto:security@larsartmann.com\""
    hostname = "@"
    mx_pref  = 10
    ttl      = 1799
    type     = "CAA"
  }
  ```
  
  ## Error
  
│ Error: Records[18].Address "0 iodef "mailto:security@larsartmann.com"" must contain a protocol prefix for CAA iodef record
│ 
│   with namecheap_domain_records.larsartmann_com,
[...]

PR #90, opened 2024-07-11
Fix mailto support for CAA 'iodef' (ref: #73)

LarsArtmann:
#73


LarsArtmann:
@vetal2409 commits are signed off, DCO passes

LarsArtmann:
Can somebody merge this?

kurok:
I will pass it to right person - thank for contribute

vetal2409:
@LarsArtmann, Thank you for your contribution. I am very sorry for the late reply. We will work on proper notification.

Changing "vendor" folder isn't a good idea because it is auto-generated and all the manual changes will be lost.
I made this fix in [go-namecheap-sdk](https://github.com/namecheap/go-namecheap-sdk) and updated this terrafrom provider with this new SDK.

[Terraform Provider v2.2.0](https://registry.terraform.io/providers/namecheap/namecheap/latest/docs) should work as you expected.  Just let us know if you have any issues with it.

PR #143, opened 2026-04-20
ci: install terraform on self-hosted runner

kurok:
## Summary

- Add a **Set up Terraform** step to the `acceptance_test` job that installs Terraform 1.14.8 into a workspace-local `.terraform-bin/` and exposes it via `TF_ACC_TERRAFORM_PATH`, so `make testacc` (terraform-plugin-sdk acceptance harness) can find a real `terraform` binary on the runner.
- Harden the existing **Set up Go** step: `curl -fsSL` so a failed download actually fails the step, and build `go-env.sh` via a `{ … } > file` block to stop double-expanding `$PATH` and to quote the workspace path.

## Why manual install instead of `hashicorp/setup-terraform`

`hashicorp/setup-terraform@v4` runs on Node 24. The self-hosted AL2023 EC2 runner used by this job doesn't support Node 24 yet — same reason `actions/checkout` is pinned to v4 in this file (see the inline comment on line 92). Pulling the binary directly keeps the job on the same Node-version floor as the rest of the steps.

## Test plan

- [x] Push triggers CI; `Set up Terraform` step prints `Terraform v1.14.8`.
- [x] `acceptance_test` job no longer fails at the terraform-lookup stage.
- [x] `go-env.sh` after the steps contains a single well-quoted `PATH=` line (no recursive `$PATH` expansion artifacts).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,18 +93,33 @@ jobs:
 
       - name: Set up Go
         run: |
-          curl https://dl.google.com/go/go1.25.9.linux-amd64.tar.gz -o go1.25.9.linux-amd64.tar.gz
+          curl -fsSLo go1.25.9.linux-amd64.tar.gz https://dl.google.com/go/go1.25.9.linux-amd64.tar.gz
           rm -rf .go-instance
           mkdir .go-instance
           tar -C .go-instance -xzf go1.25.9.linux-amd64.tar.gz
-          echo "#!/bin/bash" >> go-env.sh
-          echo "" >> go-env.sh
-          echo "export PATH=$PATH:$(pwd)/.go-instance/go/bin" >> go-env.sh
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -93,19 +93,34 @@ jobs:
...
+            echo 'export HOME="'"$(pwd)"'"'
+          } > go-env.sh
           chmod +x go-env.sh
           source go-env.sh
           go version
           go env GOPATH
 
+      - name: Set up Terraform
+        run: |
+          curl -fsSLo terraform.zip https://releases.hashicorp.com/terraform/1.14.8/terraform_1.14.8_linux_amd64.zip
```
dm1tr1yvovk, line 113: is there any specific reason to install 1.14.8 version and not newer version?
just to understand why this version was choosen

dm1tr1yvovk (approved):

PR #182, opened 2026-04-21
ci: rotate ec2-github-runner SHA to Phase 4 tip (non-root + --ephemeral)

kurok:
Dogfood rotation for [namecheap/ec2-github-runner#18](https://github.com/namecheap/ec2-github-runner/pull/18) (Phase 4 of plan #173-equivalent). Rotates both pins `a1bd2f9 → 7b949a3`.

Phase 4 is the flagged-medium-risk change: runner now runs as a dedicated non-root `runner` user with `--ephemeral` registration. A green acceptance test here confirms that `make testacc` plus the workspace-local Go/Terraform setup still work when the runner isn't root. If something fails, I'll investigate before stacking Phases 5/6/7 on top.

Issue #19, opened 2019-05-16
How do I import an already created/setup domain?


MostHated:
Hi there,
   I am just wondering how I go about configuring an already existing domain? I purchase them through Namecheap often so it already exists there, but I wanted to use Terraform to set the name servers to different things using just the "basic DNS" which provides nameservers only and I point those to Cloudflare, GCP, AWS, DigitalOcean, etc depending on various things. I tried to use namecheap_domain so I could try to import, but it said that didn't exist. Any suggestions on what I do in this situation?

Thanks,
-MH

adamdecaf:
Importing isn't supported by the provider (yet), but can be added. The Terraform `schema.Resource` struct offers an `Importer` field to handle that logic. I can try and add something in the next day or two.

https://www.terraform.io/docs/extend/resources.html#importers
https://godoc.org/github.com/hashicorp/terraform/helper/schema#Resource

MostHated:
That would be great. I always just buy them through Namecheap but never actually use their settings, I always just change the NS or something else. I have a few dozen domains there, so it would be super useful to be able to import them all and be able to mass change their records all at once. Huge time saver.  I definitely appreciate the work you have put into this. I only recently started using TF for CloudFlare, so when I came across this one, I was really excited about the amount of time I could save between the two. : D

MostHated:
Whenever you happen to have the time to make the update, if you would not mind sending a quick reply here so I know, or just close this, I would be most grateful! 

Thanks again,
-MH

adamdecaf:
I think I've got a PR for this. It works on one of my domains (root A record and subdomain), but I'd like others to try it.  Can y'all pull that branch down, or would a pre-release binary be easier? 

https://github.com/adamdecaf/terraform-provider-namecheap/pull/21

MostHated:
I can snag the branch and have a go at it. Is it 11.x or 12.x?

adamdecaf:
It's on Terraform 0.12.0

MostHated:
Never mind, it would be easier if you just made a build, lol. I am not getting import errors on the package saying there is a circular dependency.

adamdecaf:
@MostHated Sure thing, try out 1.4.0-rc1 

https://github.com/adamdecaf/terraform-provider-namecheap/releases/tag/1.4.0-rc1

MostHated:
[...]

PR #183, opened 2026-04-21
ci: rotate ec2-github-runner SHA to Phase 4 safe tip

kurok:
Second attempt at the Phase 4 dogfood after #182 failed. The action-repo [PR #19](https://github.com/namecheap/ec2-github-runner/pull/19) reverted the non-root user transition that broke registration and kept everything else from Phase 4. This pin rotation (`a1bd2f9 → 78f98d1`) is the verification.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,7 +140,7 @@ jobs:
         # SHA-pinned (was @feat/al2023-support). The same SHA is reused by
         # the stop-runner step so both halves of the runner lifecycle run
         # identical action code.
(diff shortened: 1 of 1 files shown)


kurok:
Closing — Phase 4 'safe' also failed registration timeout. The breaker is in one of the five Phase-4 changes kept after the non-root revert, not in the non-root move itself.

Full rollback of Phase 4 up on action repo at [#21](https://github.com/namecheap/ec2-github-runner/pull/21) — restores Phase 1 bootstrap byte-for-byte. Fresh dogfood PR will open against that tip once #21 merges.

Phase 4 work moves to individual small PRs, one change per dogfood, so the next failure isolates itself.

PR #208, opened 2026-05-28
chore: add -failfast to testacc to stop on first failure

kurok:
## Summary

Adds `-failfast` to `make testacc` so the acceptance test suite stops at the first failing test instead of running all remaining tests.

## Motivation

A recent failing run took **1073 s** instead of the normal ~152 s. The first ns test failed at ~100 s, but the suite continued through three more ns tests (each hitting the same timeout) plus `TestAccDomainValidation` (224 s) — all wasted time pointing at the same root cause. With `-failfast` that run would have stopped at ~100 s.

## Trade-off

On a genuine regression affecting multiple tests you only see the first failure per run. Given the suite is small and sequential (single domain), discovering failures one-at-a-time is acceptable.

## Test plan

- [ ] CI passes on this PR
- [ ] Confirm `make testacc` locally exits immediately on first failure when one test is broken

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ test:
 # Please set the following ENV variables for this test:
 # NAMECHEAP_USER_NAME, NAMECHEAP_API_USER, NAMECHEAP_API_KEY, NAMECHEAP_TEST_DOMAIN, NAMECHEAP_USE_SANDBOX (optional, default is false)
 testacc:
-	TF_ACC=1 go test ./namecheap -v -run=TestAcc -count=1 -timeout=30m
+	TF_ACC=1 go test ./namecheap -v -run=TestAcc -count=1 -timeout=30m -failfast
 
 build:
 	go build -o ${BINARY}


Makefile
```diff
@@ -15,7 +15,7 @@ test:
 # Please set the following ENV variables for this test:
 # NAMECHEAP_USER_NAME, NAMECHEAP_API_USER, NAMECHEAP_API_KEY, NAMECHEAP_TEST_DOMAIN, NAMECHEAP_USE_SANDBOX (optional, default is false)
 testacc:
-	TF_ACC=1 go test ./namecheap -v -run=TestAcc -count=1 -timeout=30m
+	TF_ACC=1 go test ./namecheap -v -run=TestAcc -count=1 -timeout=30m -failfast
```
dm1tr1yvovk, line 18: I have doubts about it
running all tests would show full picture and all possible failures to fix in one go
to see all surface area of failures, not only one first failed test.
having tests to run 2-3 minutes(when our API works fine) is not very long

kurok: The case for it: the slow 1073s run we saw was almost entirely wasted time — the first ns test failed at ~100 s, but the suite kept running 3 more ns tests (each hitting the same timeout) plus TestAccDomainValidation (224s). With -failfast, that run would have stopped at ~100 s instead of 1073 s.
The suite is small enough that discovering failures one at a time is fine, and the time saved on flaky/broken runs is significant.

dm1tr1yvovk: seeing all failed tests shows us that now only one test is not working, but the exact functionality in our API and seeing it fully even longer could be considered acceptable.
but you fixed tests with timeouts already, so we would not wait for 1000 seconds anymore.

kurok: I want faster feedback - it is rare to have API broken - but fail with changes in provider more important to se faster

dm1tr1yvovk (approved):

PR #211, opened 2026-05-29
ci: update go version in workflows

dm1tr1yvovk:


kurok (approved):
The change itself is correct — `go-version-file: go.mod` is the right pattern and aligns `release.yml` with how the `check` job already works.

One related issue worth addressing:

**`ci.yml` `acceptance_test` — Go version pin is now dead**

The `acceptance_test` job manually downloads `go1.25.9` via curl, but `go.mod` requires `go 1.26.3`. With `GOTOOLCHAIN=auto` (the Go default since 1.21), the first `go` invocation will attempt to auto-download `go1.26.3` from `dl.google.com`, silently bypassing the pinned install. On the self-hosted EC2 runner, if outbound access to `dl.google.com` is restricted, the acceptance job will hard-fail at the first `go` call with no obvious error.

The fix: replace the hand-rolled curl install in `acceptance_test` with `actions/setup-go@v6` + `go-version-file: go.mod`, matching the `check` job pattern. This also ensures future `go.mod` bumps automatically propagate to all three jobs without a manual curl step update.

PR #258, opened 2026-07-02
chore(deps): adopt go-namecheap-sdk v2.7.0 (migrate DNS calls to WithContext API)

kurok:
## Summary

Adopts **go-namecheap-sdk `v2.7.0`** (up from `v2.6.4`) and migrates the provider to the SDK's context-aware DNS API.

`v2.7.0` deprecates the non-context `DomainsDNS`/`Domains` methods (`GetList`, `GetHosts`, `SetHosts`, `SetCustom`, `SetDefault`) in favour of their `*WithContext` variants. Building against it without migrating fails `golangci-lint` (`staticcheck SA1019`), so this PR switches every call site to the context-aware methods.

## Changes

- **`go.mod` / `go.sum`**: `github.com/namecheap/go-namecheap-sdk/v2` `v2.6.4` → `v2.7.0`.
- **`namecheap/namecheap_domain_record.go`**, **`namecheap_domain_record_functions.go`**: replace deprecated calls with `*WithContext`, threading the resource's `context.Context` from the CRUD entry points (`resourceRecordCreate/Read/Update/Delete`) down through the 14 DNS/nameserver helper functions.
- **Tests** (`*_crud_*_test.go`, `namecheap_domain_record_test.go`, `provider_test.go`): pass `context.Background()` to the migrated helpers/SDK calls.

## Behaviour / compatibility

No functional change for users — **no plan or resource diffs**. The deprecated SDK methods were thin wrappers that already called `…WithContext(context.Background(), …)`; this only threads the real request context through. No business logic, output, or provider schema changes.

## Testing

- `go build ./...`, `go vet ./...` — clean
- `golangci-lint run` — **0 issues** (SA1019 resolved)
- `go test ./namecheap/` — pass
- CI on the branch is green, including the **acceptance test** suite (run 28610058245).


dm1tr1yvovk:
### Code review

The SDK v2.7.0 / `WithContext` migration itself is clean and mechanical: `context` is threaded from all four CRUD entry points through the helper functions, every deprecated call has a `WithContext` replacement, and the prior nil-guard / error-handling fixes are preserved. No blocking issues in the migration.

Two follow-ups:

**1. The two new `email_type` tests are vacuous — please strengthen them.**

Both tests pre-seed `email_type = "MX"` into the resource data *before* calling `resourceRecordRead`, then assert it is still `"MX"`. Since production code only writes it back when `emailType != nil`, the assertion is satisfied by the pre-seeded value regardless of whether the read logic runs — the tests would still pass if the `data.Set("email_type", ...)` write-back were deleted.

Overwrite test:

https://github.com/namecheap/terraform-provider-namecheap/blob/7b49641833990aaac17706c1520d6dde957ed5fa/namecheap/namecheap_domain_record_crud_handler_test.go#L659-L668

Merge test:

https://github.com/namecheap/terraform-provider-namecheap/blob/7b49641833990aaac17706c1520d6dde957ed5fa/namecheap/namecheap_domain_record_crud_handler_test.go#L688-L697

Could you add coverage for the case that actually exercises the write-back — `email_type` absent from state (`emailType == nil`) but present in the API response (the `terraform import` path)? That verifies the value flows from the remote response into state rather than merely surviving.

**2. Pre-existing bugs surfaced during review (not caused by this PR).**

I filed these separately so this PR stays a clean mechanical migration:

- #260 — `readRecordsOverwrite` does not filter default parking records, causing OVERWRITE-mode drift on parked/imported domains.
- #261 — `validateGetHostsResponse` ignores `response.Errors`, the documented `GetHosts` error-check gotcha.

Nothing here blocks the migration; #1 is the only ask for this PR.

kurok:
Addressed in 918b833.

**Strengthened tests:** both tests now pre-seed `email_type = "FWD"` while the mocked API returns `"MX"`, and assert state ends up as `"MX"`. Verified by mutation: with the `data.Set("email_type", ...)` write-back deleted, both tests fail (`FWD` != `MX`), so they now exercise the write-back rather than the pre-seeded value surviving.

**On the suggested import-path case:** it does not exercise the write-back under the current code. When `email_type` is absent from state, `GetOk` returns false, `emailType == nil`, and the write-back is skipped — so on `terraform import` the value never flows from the API response into state. Confirmed empirically before writing the test. The guard predates this PR (it avoids injecting `email_type` into state for configs that never set it, which would otherwise produce a perpetual plan diff). I added `TestResourceRecordRead_ImportWithoutEmailTypeInState` pinning that current behavior: import leaves `email_type` unset even when the API reports `MX`. If import *should* capture the remote `email_type`, that is a behavior change — probably best filed alongside #260/#261 rather than folded into this migration.

dm1tr1yvovk (approved):

Issue #62, opened 2021-12-05
Allow IPs for Terraform Cloud


cvillers:
Terraform Cloud must be allowed API access. Currently, I have to manually allow each IP address so that applying via Terraform Cloud succeeds. From experimentation, it seems that different IPs are used for different runs, so I have to add a new one after each failure.

A related issue is #59 - if there is a public CIDR for cloud-apply that could also work here. Note that the [IP Range API](https://www.terraform.io/docs/cloud/api/ip-ranges.html) does not return the cloud-apply ranges so I can't predict ahead of time what to use. 

### Terraform Version
1.0.11

### Namecheap provider version
2.0.2

### Expected Behavior
I should be able to use Terraform Cloud as my backend and perform an apply from there.

### Actual Behavior
```
$ terraform apply

╷
│ Error: Invalid request IP: 3.145.98.94 (1011150)
│ 
│   with namecheap_domain_records.MY_DOMAIN,
│   on domain.tf line 3, in resource "namecheap_domain_records" "MY_DOMAIN":
│    3: resource "namecheap_domain_records" "MY_DOMAIN" {
│ 
╵
```

### Steps to Reproduce
1. Use Terraform Cloud (the service, not self-hosted) as the backend.
2. `terraform apply`

cvillers:
After some experimentation, I found that 2 IPs that have failed so far are part of AWS us-east-2, and another is in us-east-1. I understand that it is probably undesirable to allow entire AWS CIDR ranges so #59 may not solve this after all.

adamdecaf:
I don't know the Namecheap internals but their systems might take a few minutes to fully update each allowed IP.

StyleT:
Hi!
Unfortunately this is a limitation of the Namecheap public API itself, rather than Terraform provider itself. Also there is no way to change whitelist via API as far as I know :(

I would advice you to contact our customer support team. They do track such client requests and as soon as there is enough demand on certain feature - it get's processed by respective product team.

tedmiston:
This seems to make the Namecheap provider totally unusable when Terraform Cloud is the backend.  I keep whitelisting new IPs, but from the TFC side it seems that the IPs are dynamic, so it fails every time.

It's pretty hard to do anything real with this provider if the underlying API breaks on requiring whitelisting something that Terraform Cloud doesn't have (static IPs).

Lumexralph:
[...]

PR #281, opened 2026-07-04
ci: warm-pool EC2 runner lifecycle with cross-repo EIP mutex and per-job hygiene sweeps

### LINKED ISSUE #279: CI: warm-pool EC2 runner lifecycle — stop/reuse between runs, daily terminate, per-job hygiene, cross-repo single-flight gate on the sandbox EIP
## Motivation

Every push today cold-launches a fresh EC2 runner for the sandbox acceptance suite (instance launch + OS boot + yum + ~100 MB runner download + Go/Terraform install ≈ 2–4 min of boot tax per run) and terminates it afterwards. [`namecheap/ec2-github-runner` v4.0.0](https://github.com/namecheap/ec2-github-runner/releases/tag/v4.0.0) shipped warm pools (`reuse: stop`), a scheduled reaper (`mode: cleanup` + `reaper-stopped-max-age`), a TTL self-destruct (`max-lifetime-minutes`) and `pre-runner-script` — everything needed to switch to a **stop/reuse lifecycle**:

- **stop, don't terminate** after each run — a stopped t3.medium costs only EBS and restarts in seconds with its disk intact;
- **terminate every day** — the first acceptance run of the day cold-launches a fresh instance, a nightly reaper drains the pool, so no instance ever lives past ~24 h;
- **per-job hygiene** — repo code, env files and anything secret-adjacent live in an ephemeral per-job workspace that is wiped before and after every job; only *prerequisites* (Go toolchain, Terraform binary, build/module caches) persist across the stop/start cycle;
- **single-flight gate across repos** — only one CI job anywhere may hold the whitelisted sandbox EIP at a time.

## Facts that shape the design (verified against ec2-github-runner v4.0.0 source)

1. **Warm pools are per-repository.** `findStoppedPoolInstance` filters on the `ec2-github-runner:repository` signature tag, and runners register with `config.sh --url github.com/<owner>/<this-repo>` — so another repo's CI can *never* schedule a job onto this repo's instance. The genuinely shared, contended resource is the **whitelisted Elastic IP**, not the instance. The cross-repo gate therefore has to guard the EIP.
2. **Warm start does not (re-)associate the EIP.** Only the cold-launch path calls `AssociateAddress`. A VPC EIP normally stays attached across stop/start — which means a stopped warm instance would silently monopolize the EIP 24/7 and starve any other repo. The plan manages the EIP explicitly in the workflow as an acquire/release mutex instead (see item 3 below).
3. **`AssociateAddress` without `AllowReassociation` is an atomic test-and-set.** It fails if the EIP is attached elsewhere — a free, infrastructure-less distributed lock. (Note the action itself only *warns* and proceeds without the EIP on association failure, so the action's built-in `eip-allocation-id` input alone is not a hard gate.)
4. **The TTL self-destruct arms only on cold boot.** `shutdown -h +N` is injected via user-data, which cloud-init runs once per instance; the pending timer dies when the instance stops and is not re-armed on warm boots. Daily termination must therefore be enforced by the `mode: cleanup` reaper (`reaper-stopped-max-age`), with the TTL as a first-session backstop only. *(Re-verify this during implementation with a live warm cycle.)*
5. **Warm reuse means job N+1 runs on job N's disk.** The README is explicit that this is only safe for a single trusted repo's CI. Our sandbox pipeline is already push-only (fork PRs are served by `acceptance_mock`); that guard becomes **load-bearing for warm-pool safety** and gets a comment saying so.

### Lifecycle

```
first push of the day        subsequent pushes                nightly (cron)
──────────────────────       ─────────────────────            ──────────────────────
cold launch (2–4 min)   ──▶  warm start (~seconds)   ──▶ ...  mode: cleanup reaper
join pool, run job           run job                          terminates stopped
stop + release EIP           stop + release EIP               pool instance(s)
```

### Harness gates (summary)

| # | Gate | Mechanism | Status |
|---|------|-----------|--------|
| G1 | Sandbox pipeline never runs fork/PR code | `if: github.event_name == 'push'` on start-runner / acceptance / stop-runner | existing — now load-bearing for warm-pool safety |
| G2 | No runner before checks pass | `needs: [check, security, govulncheck]` | existing |
| G3 | Single-flight within this repo (FIFO) | `ahmadnassri/action-workflow-queue` | existing |
| G4 | Single-flight **across repos** | EIP as atomic mutex: acquire with `--no-allow-reassociation` retry loop, release on stop | **new** |
| G5 | Tests never run without the whitelisted IP | hard assert `public IP == EIP` before `make testacc` | **new** |
| G6 | No code/env/secrets survive on the stopped disk | pre-job + post-job hygiene sweeps; persistent-path allowlist | **new** |
| G7 | Instance never lives past the day | nightly `mode: cleanup` drain + `max-lifetime-minutes` first-session TTL + `reuse-max-cycles` | **new** |

### 1. Switch the runner to a warm pool (`.github/workflows/ci.yml`)

- `start-runner` / `stop-runner` steps: add `reuse: stop` (both sides, required by the action) and `reuse-pool-tag: sandbox-eip` (one pool tier — the single EIP means effective pool size 1).
- Keep `max-lifetime-minutes` at the default 360 (first-session backstop, well above the longest acceptance run); keep `reuse-max-cycles` default (20) — a mid-day recycle just cold-launches back into the pool.
- Keep the existing `ahmadnassri/action-workflow-queue` step (G3) — the warm pool holds one instance, so in-repo FIFO serialization still prevents two runs fighting over it.

### 2. EIP as an explicit cross-repo mutex (G4, G5)

Remove `eip-allocation-id` from the action inputs and manage the EIP in workflow steps (required anyway, because warm starts never re-associate — fact 2):

- **Acquire** (in `start-runner`, after `mode: start`): retry loop over
  `aws ec2 associate-address --allocation-id eipalloc-1796f61b --instance-id <id> --no-allow-reassociation`
  — atomically fails while any other instance (any repo) holds the IP; poll with backoff, hard timeout ~30 min to match the queue gate. Re-associating to the *same* instance succeeds, so a crashed previous run in this repo self-heals.
- **Verify** (first step of `acceptance_test`): assert the instance's public IP equals the EIP; fail fast otherwise. This turns "the action only warns on association failure" into a hard gate.
- **Release** (in `stop-runner`, `if: always()`, after `mode: stop`): `aws ec2 disassociate-address` — the stopped instance keeps its disk/caches but frees the IP for other repos' CIs.
- Leak path: if `stop-runner` never runs, the EIP stays held until the reaper terminates the holder (termination auto-disassociates). The cleanup workflow (item 4) also gets a belt-and-braces step: if the EIP is associated to a *stopped* instance, disassociate it.
- IAM: the CI user needs `ec2:AssociateAddress`, `ec2:DisassociateAddress`, `ec2:DescribeAddresses` (the first was already exercised by the action under the same credentials; confirm all three).
- Cross-repo fairness is best-effort (retry loop, not FIFO); in-repo FIFO is preserved by G3. Acceptable at current contention levels.

### 3. Per-job ephemeral workspace + persistent prerequisite layer (G6)

Split the disk into an explicit allowlist:

**Persists across stop/start (prerequisites only — wiped by the daily terminate):**
- `/opt/ci/toolcache/go-<version>/`, `/opt/ci/toolcache/terraform-<version>/` — SHA-verified installs keyed by full version, so the existing "latest patch of the pinned minor" resolution keeps working (a patch bump installs alongside, no drift);
- `/opt/ci/cache/go-build/`, `/opt/ci/cache/go-mod/` — `GOCACHE`/`GOMODCACHE`;
- `/home/runner/actions-runner/` — the runner itself (action-managed).

**Ephemeral (wiped before AND after every job):**
- `/home/runner/actions-runner/_work/` — checkout, `go-env.sh`, Terraform working dirs, test artifacts;
- `/opt/ci/jobs/<run-id>/` — per-job `$HOME`, so stray dotfiles (`~/.terraform.d`, `~/.config`, rc files) never land on the persistent layer;
- stale runner `_diag` logs.

Job restructure in `acceptance_test`:
- **Step "hygiene sweep (pre)"**: wipe ephemeral paths; if leftovers from a previous job are found, emit a `::warning` (canary that a cancelled/crashed run skipped its post-sweep — the pre-sweep is what makes the cancel path safe, since a killed runner never executes its `always()` steps).
- Replace `actions/cache` (network restore into the workspace) with the local persistent caches — faster on warm starts and removes a cache-poisoning surface.
- Convert the Go/Terraform install steps to "ensure in toolcache" (skip download when the resolved version is already present); keep the existing version-resolution and sha256 pinning logic verbatim.
- **Step "hygiene sweep (post)"** with `if: always()`: wipe the ephemeral paths again.
- `NAMECHEAP_API_KEY` stays step-level env only (process memory, dies with the job); nothing may write it under a persistent path.
- `pre-runner-script` (new v4 input) on the start step provisions `/opt/ci/{toolcache,cache,jobs}` with runner-user ownership on cold boot.

Net effect: a stopped instance's disk holds toolchains and build caches, but no repo code, no env files, no secrets — "ready to work with" for the next run, clean if anyone inspects it.

### 4. Daily lifecycle: nightly drain + leak reaper (G7)

New workflow `.github/workflows/cleanup-ec2-runners.yml` (based on the action's [docs/cleanup-workflow.yml](https://github.com/namecheap/ec2-github-runner/blob/main/docs/cleanup-workflow.yml)), one job, two cron schedules branched on `github.event.schedule`:

- **Nightly drain** — `cron: '30 2 * * *'` (UTC): `mode: cleanup` with `reaper-stopped-max-age: 30` and `max-age-minutes: 30` → terminates the day's stopped pool instance(s) and any idle stragglers. Next morning's first push cold-launches fresh — exactly the "terminate completely at end of day / fresh instance on first run" requirement. Busy runners are never reaped, so a late-night in-flight job is safe.
- **Leak reaper** — `cron: '*/30 * * * *'`: `mode: cleanup` with defaults (`max-age-minutes: 120`, `reaper-stopped-max-age: 1440`) → catches crashed/cancelled leftovers during the day *without* draining the warm pool.
- `workflow_dispatch` with a `dry-run` input (the reaper prints a job-summary table of examined/reaped/skipped).
- Extra step (see item 2): release the EIP if it is found associated to a stopped instance.
- Uses the same GitHub App token + AWS credentials as CI; reaper needs only `ec2:DescribeInstances` + `ec2:TerminateInstances` on the AWS side (already granted).

### 5. Docs & comments

- Update the workflow header comments: the push-only guard is now also the warm-pool safety precondition (fact 5); document the EIP-mutex protocol so other repos that want sandbox access know the contract (acquire with `--no-allow-reassociation`, verify, release on stop).
- Note in CONTRIBUTING/README how to manually drain the pool (`workflow_dispatch` the cleanup workflow) and how to debug a failed start (`cleanup-on-start-failure: false`).

### 6. Optional upstream follow-up (namecheap/ec2-github-runner)

The workflow-level EIP steps in item 2 exist because the action's EIP handling doesn't cover the warm-pool lifecycle. Propose three small inputs upstream, after which items 2's custom steps collapse back into action inputs:
- re-associate `eip-allocation-id` on warm start;
- `eip-strict: true` — fail the start (instead of warn) when association fails;
- `eip-release-on-stop: true` — disassociate in `mode: stop`.

## Verification plan

1. Two consecutive pushes: second run's start log shows `Reused warm-pool EC2 instance`, start latency drops from minutes to seconds.
2. EIP contention: manually associate the EIP elsewhere, push — the acquire step waits/timeouts and G5 proves no test ran without the whitelisted IP.
3. Hygiene: after a normal run, inspect the stopped instance via SSM — no `_work` content, no `go-env.sh`, no job `$HOME`; grep the persistent layer for the API key finds nothing.
4. Cancel a run mid-test: next run's pre-sweep canary warns and cleans up; job still passes.
5. Nightly: next morning `describe-instances` shows no pool instance; first push cold-launches; the drain run's job-summary table shows the reap reason `stopped-past-max-age`.
6. Reaper `dry-run: true` via `workflow_dispatch` before enabling the schedules.

## Risks & mitigations

- **State carry-over between jobs** (warm disk): trusted push-only pipeline only (G1), hygiene sweeps (G6), 24 h cap (G7), `reuse-max-cycles` recycle.
- **TTL not re-armed on warm boots** (fact 4): the reaper is the authoritative bound; TTL is only a first-session backstop. Verify live during implementation.
- **Cancelled run skips `always()` cleanup**: pre-job sweep + canary covers it.
- **Toolchain drift within the day**: version resolution runs every job; new patches install alongside in the toolcache; daily terminate bounds everything at 24 h.
- **Other repos starved of the EIP**: release-on-stop (item 2) frees it between runs — strictly better than today, where the instance holds the EIP for the whole run and the queue is invisible cross-repo.

## Out of scope

- Sharing one runner instance across repos (impossible: runner registration and pool signature tags are per-repo by design).
- Changes to the mock acceptance matrix, Trivy/govulncheck gates, or the release pipeline.
- The upstream action changes (item 6) — separate epic in namecheap/ec2-github-runner if we want them.

Related: #246 (acceptance-test infrastructure epic — this issue covers the runner lifecycle underneath the gated sandbox suite described there).

kurok:
## Summary

Reworks the `acceptance_test` self-hosted EC2 runner lifecycle from
terminate-every-run to a warm pool (`reuse: stop` / `reuse-pool-tag`), so most
pushes warm-start in seconds instead of paying a ~2-4 minute cold boot every
time. To make that safe, this PR also adds:

- **`scripts/eip-mutex.sh`** — replaces the ec2-github-runner action's
  permissive `eip-allocation-id` input (which reassociates the whitelisted
  sandbox Elastic IP unconditionally) with an atomic
  `associate-address --no-allow-reassociation` test-and-set. `acquire` retries
  only on `Resource.AlreadyAssociated` contention (bounded, 1800s default);
  `release` now requires the instance-id this job itself started and refuses
  to disassociate if the EIP has since been reassociated to a *different*
  instance; `reap-if-stopped` is a belt-and-braces cleanup that only acts on
  an EIP held by a genuinely stopped instance.
- **`scripts/hygiene-sweep.sh`** — wipes the per-job `$HOME`
  (`/opt/ci/jobs/current`, including a job-scoped `TMPDIR` so Go/
  terraform-plugin-testing scratch files don't leak) both before (self-heals +
  warns on a crashed/cancelled prior run's leftovers) and after
  (`if: always()`) every acceptance-test job, since warm reuse means job N+1
  runs on job N's disk.
- **`.github/workflows/cleanup-ec2-runners.yml`** — a new scheduled workflow
  with two passes: a nightly full drain (guarantees the pool is terminated at
  least once a day) and a more frequent leak reaper (catches crashed/cancelled
  leftovers at the action's default threshold). Both passes also run the
  EIP's `reap-if-stopped` belt-and-braces check. `workflow_dispatch` exposes
  `mode` and a `dry_run` (default `true`) input for manual/preview runs.
- `SECURITY_COMPLIANCE.md` and `CONTRIBUTING.md` updated to document the new
  warm-pool lifecycle, hygiene sweeps, EIP mutex, and the new IAM
  prerequisite below.

Closes #279.

## Manual AWS IAM prerequisite — this PR cannot be merged without it

The CI AWS identity currently has `ec2:AssociateAddress` and
`ec2:DescribeInstances`, but the new mutex additionally requires:

- [ ] `ec2:DisassociateAddress` granted to the CI role/user
- [ ] `ec2:DescribeAddresses` granted to the CI role/user

**This repo does not manage that IAM policy — an AWS admin must grant both
actions before this pipeline will work.** Until that happens, expect the
`acceptance_test` run on this PR to fail with `AccessDenied` at the
`Acquire sandbox EIP` and/or `Release sandbox EIP` steps. That failure mode is
expected and diagnostic — confirmation the mutex code path is being
exercised and correctly surfacing the missing grant — not a bug in this
PR's code. Please don't debug the workflow logic in response to it; get the
grant applied and re-run instead.

## Cross-repo hazard: `namecheap/mcp-server-namecheap` shares this EIP

The whitelisted sandbox Elastic IP (`eipalloc-1796f61b`) is also used by
`namecheap/mcp-server-namecheap`'s acceptance workflow, which as of this
writing still passes it straight to `ec2-github-runner`'s permissive
`eip-allocation-id` input (allows reassociation) rather than anything like
the mutex added here. Concretely: that workflow can silently steal the EIP
out from under a concurrently-running job in *this* repo, and prior to this
PR's `cmd_release` fix, this repo's own release step could have done the same
back to them.

This PR makes **this repository** a well-behaved actor — it will wait on, or
loudly fail on, contention rather than silently stealing the address, and it
now verifies ownership before releasing so it won't rip the EIP away from a
different instance that reassociated it mid-run. It **cannot**, however,
force `mcp-server-namecheap` to adopt the same discipline; as long as that
workflow keeps using the permissive path, occasional `acceptance_test`
failures at the "Verify sandbox EIP" step in *either* repo are possible and
are most likely caused by the other repo's concurrent run, not a bug here.
Raising this with `mcp-server-namecheap`'s maintainers (e.g. filing a
companion issue asking them to adopt an equivalent acquire/release mutex) is
left to a human to do separately — out of scope for this PR, which touches
only this repository.

## Adversarial review: confirmed findings, fixed

All 9 confirmed findings from adversarial review were fixed in this PR
(none were skipped):

1. New `actions/checkout@v7` steps in the secret-bearing start-runner/
   stop-runner jobs and the new cleanup reap job were on a floating tag
   instead of the repo's required 40-char SHA — pinned to the verified
   `actions/checkout` `v7.0.0` SHA (`9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0`).
   Pre-existing floating `@v7` usages elsewhere in `ci.yml` are untouched —
   out of scope for this PR.
2. `hygiene-sweep.sh` never cleared `TMPDIR`/`/tmp`, so Go/
   terraform-plugin-testing scratch files could survive on the reused disk —
   `TMPDIR` is now pointed at a subdirectory of the already-swept job-home
   path.
3. The new `cleanup-ec2-runners.yml` `reap` job requested an unused
   `actions: read` permission — removed; only `contents: read` remains.
4. `eip-mutex.sh release` disassociated whatever was currently associated
   without checking it belonged to this job's own instance, enabling a
   cross-repo steal-then-strand — `release` now takes the job's own
   instance-id and no-ops if the current association doesn't match.
5. `hygiene-sweep.sh`'s post-sweep could `rm -rf` the invoking shell's own
   cwd, leaving `$PWD` pointed at a deleted inode after recreation — it now
   steps out to `/` first and back in afterward when that's the case.
6. `resolve_association_id`/`resolve_associated_instance_id` swallowed *all*
   `describe-addresses` failures (not just the legitimate "unassociated"
   case) via `2>/dev/null || true`, which could mask a real IAM/AWS error as
   "nothing to release" — the swallow was removed; genuine failures now
   propagate under `set -e`.
7. The new stop-runner checkout step lacked `if: always()`, so an earlier
   token/credentials failure would skip it and break the `always()`-guarded
   EIP release step downstream — added.
8. The acceptance_test checkout step (already first-in-job, so `if: always()`
   would be a no-op) had the same underlying risk for the post-job hygiene
   sweep — fixed at the actual failure point instead: the post-job sweep step
   now falls back to inline cleanup with a `::warning::` if
   `scripts/hygiene-sweep.sh` isn't on disk.
9. A few new `${{ steps.*.outputs.* }}` values were spliced directly into
   `run:` shell text instead of routed through `env:`, inconsistent with the
   safer pattern already used two lines away — routed through `env:`
   throughout.

## Why draft

Per this repo's `CONTRIBUTING.md` convention of submitting WIP PRs: this is
explicitly WIP pending the IAM grant above. `acceptance_test` on this PR is
expected to fail until an AWS admin applies it.


kurok:
## Live sandbox run result: reproducible cold-boot registration failure under `reuse: stop`

Two consecutive pushes to this branch both failed at the pre-existing **"Start EC2 runner"** step — before any of this PR's own EIP-mutex/hygiene code ever ran:

| Run | Instance | Result |
|---|---|---|
| [28718108647](https://github.com/namecheap/terraform-provider-namecheap/actions/runs/28718108647/job/85163000709) | `i-0849411016f0aaceb` | Registration timeout (5 min), no console output captured, auto-terminated |
| [28718191197](https://github.com/namecheap/terraform-provider-namecheap/actions/runs/28718191197/job/85163205523) | `i-0ec6599981ff8c88d` | Same: registration timeout (5 min), no console output captured, auto-terminated |

Both runs used identical `ec2-image-filters`, `ec2-instance-type`, `subnet-id`, `security-group-id`, and `iam-role-name` to the pipeline that's been working on `master` — the only change to the `start-ec2-runner` step in this PR is removing `eip-allocation-id` and adding `reuse: stop` + `reuse-pool-tag: sandbox-acceptance`.

Since this pool tag has never been used before, both attempts necessarily went through the **cold-launch warm-pool bootstrap path** (`buildReusableUserData` in `namecheap/ec2-github-runner`, shipped 2 days ago in v4.0.0) rather than the plain one-shot bootstrap the current pipeline has always used. Two independent instances failing identically, with **zero console output captured in either case** (the action's own diagnostics say "the instance may have failed before cloud-init produced output"), points at a real problem with that new bootstrap path against this AMI/environment rather than a one-off flake — but I don't have AWS console/CloudTrail access to inspect the instances directly (both were auto-terminated by the action's `cleanup-on-start-failure` behavior), and I didn't find an existing report of this in `namecheap/ec2-github-runner`'s issue tracker.

**I did not keep retrying** — each attempt costs a real AWS EC2 launch, and two identical failures is a strong enough signal without spending more.

**This blocks the acceptance_test job entirely** (upstream of the IAM-permission gap already called out above) until someone with AWS console access can either reproduce this by hand (`RunInstance` with the same warm-pool user-data and inspect the serial console directly, since GitHub's own capture found nothing) or confirm whether this AMI (`nc-al2023-ami-hvm-x86_64-master-*`) is compatible with the new warm-pool systemd-based re-registration bootstrap. Happy to iterate once there's a diagnosis — flagging this now rather than pushing further live attempts.

kurok:
Pushed `71888d3` fixing a live-test failure and a design gap in the EIP mutex from the prior commit.

**Root cause of the failure:** the acceptance runner never got a public route. CloudTrail confirmed no public IP was ever assigned to the launched instance, and this VPC has no NAT gateway — so once the previous commit deferred EIP association until *after* `mode: start` returned (via the `acquire`/`--no-allow-reassociation` test-and-set), the instance had no way out to the internet during its own bootstrap (runner tarball download, `config.sh`/`run.sh` registration).

**Architectural constraint found while fixing it:** simply restoring `eip-allocation-id` on the cold-launch `Start EC2 runner` step isn't sufficient on its own. `namecheap/ec2-github-runner`'s warm-restart path (`reuse: stop`) never re-associates the EIP itself — it assumes the EIP is already attached from a previous run. So a warm-started instance needs the EIP in place *at boot*, not just on the first cold launch.

**Confirmed design:** the EIP now stays attached to the warm-pool instance across the whole day (every stop/start cycle), instead of being acquired/released per job:
- `wait-until-free` runs as a precondition gate immediately before the cold-launch association, so it doesn't race a still-live holder (and reaps a foreign repo's stale stopped-instance association if it finds one).
- `verify` runs as a hard-fail post-check after `mode: start` returns (cold or warm), asserting the EIP actually ended up on this job's instance before `make testacc` is allowed to run.
- There's no explicit release step anywhere. The EIP is only freed as an automatic AWS side effect of `cleanup-ec2-runners.yml` terminating the pool instance — normally the nightly full drain, occasionally the leak-reaper pass if the instance ages out between drains.

This is a real trade-off, not a side effect I missed: cross-repo EIP fairness during the day is weaker than the original `acquire`/`--no-allow-reassociation` atomic test-and-set (which could only fail loudly on contention, never silently lose the address). `wait-until-free`/`verify` can only *detect* a steal after the fact and hard-fail the run, not prevent one from happening in the first place. `SECURITY_COMPLIANCE.md`'s cross-repo hazard section is updated to say this explicitly, pending `namecheap/mcp-server-namecheap` adopting the same wait/verify (or `--no-allow-reassociation`) discipline on its side.

Also folded in from review: a bug where `wait-until-free`'s reap-a-foreign-stopped-instance branch looped via `continue` without ever checking `EIP_MUTEX_TIMEOUT_SECONDS`, a `set -e` hardening on that same reap call (so a raced reclaim by the foreign owner logs-and-retries instead of hard-crashing the step), a stdout leak from `cmd_reap_if_stopped`'s status prose into `wait-until-free`'s documented exit-code-only contract, and a couple of doc corrections (the retry loops are flat fixed-interval, not "backoff"; EIP release is normally-but-not-always via the nightly drain specifically).

Leaving this as draft: the IAM prerequisite this PR's description already calls out (`ec2:DisassociateAddress` / `ec2:DescribeAddresses` for the acquire/verify/release calls) is still an open gap in production, so `verify`/`reap-if-stopped` can't be exercised for real yet.

kurok:
## Live confirmation + the complete IAM gap (not just the two actions originally flagged)

Pushing the EIP-timing fix (commit 71888d3) triggered a real run: it now progresses correctly through checkout and into the new **"Wait for sandbox EIP to become free"** step — a huge diagnosability improvement over the old silent 5-minute hang — and fails immediately (4s) with a precise error:

```
An error occurred (UnauthorizedOperation) when calling the DescribeAddresses operation:
User: arn:aws:iam::989566014062:user/sys_github_runner_provisioner is not authorized to
perform: ec2:DescribeAddresses because no identity-based policy allows the ec2:DescribeAddresses action
```

This is the expected/diagnostic blocker the PR description already calls out. I pulled `sys_github_runner_provisioner`'s current inline policy (read-only access) to get the complete picture rather than guessing further, and the gap is bigger than the two actions originally flagged:

**Missing entirely from the current policy:**
- `ec2:DescribeAddresses`, `ec2:DisassociateAddress` — the EIP mutex itself (confirmed live above).
- `ec2:StopInstances`, `ec2:StartInstances`, `ec2:ModifyInstanceAttribute` — needed by `reuse: stop` itself (stopping the instance every run, and warm-restarting/rewriting its user-data on reuse). Without these, warm-pool reuse can't work at all regardless of the EIP fix.
- `ec2:DescribeTags`, `ec2:GetConsoleOutput` — needed by `ec2-github-runner`'s own bootstrap fast-fail/console-capture diagnostics (the caller-side half; the instance's own self-tagging via `AmazonSSMRoleForInstancesQuickSetup` already works fine, confirmed via CloudTrail). This one explains a mystery from the earlier investigation: it's why my original two failed attempts got zero console output and no fast-fail — that feature has been silently degraded to timeout-only detection this whole time, for lack of this permission, not a bug in the action.

**Suggested policy addition** (matching the existing statements' resource-scoping style):

```json
{
  "Sid": "EipMutexQueries",
  "Effect": "Allow",
  "Action": ["ec2:DescribeAddresses", "ec2:DescribeTags"],
  "Resource": "*"
},
{
  "Sid": "EipMutexRelease",
  "Effect": "Allow",
  "Action": "ec2:DisassociateAddress",
  "Resource": "arn:aws:ec2:*:989566014062:elastic-ip/eipalloc-1796f61b"
},
{
  "Sid": "BootstrapDiagnosticsConsoleOutput",
  "Effect": "Allow",
  "Action": "ec2:GetConsoleOutput",
  "Resource": "arn:aws:ec2:us-west-1:989566014062:instance/*"
},
{
  "Sid": "WarmPoolLifecycle",
  "Effect": "Allow",
  "Action": ["ec2:StopInstances", "ec2:StartInstances", "ec2:ModifyInstanceAttribute"],
  "Resource": "arn:aws:ec2:us-west-1:989566014062:instance/*",
  "Condition": {
    "StringEquals": { "ec2:ResourceTag/Name": "github_runner" }
  }
}
```

(`DescribeAddresses`/`DescribeTags` don't support resource-level scoping in IAM, hence `Resource: "*"` for those two — consistent with the existing `DescribeImages`/`DescribeInstances`/`DescribeInstanceStatus` statement already in this policy.)

I don't have write access to apply this — flagging the complete, precise diff here so whoever grants it doesn't have to rediscover the remaining gaps one UnauthorizedOperation at a time.

.github/workflows/ci.yml
```diff
@@ -18,6 +18,12 @@ on:
 permissions:
   contents: read
 
+env:
+  # Single source of truth for the whitelisted sandbox Elastic IP within this
+  # file. cleanup-ec2-runners.yml repeats this same literal in its own
+  # top-level `env:` block — keep the two in sync if this value ever changes.
+  SANDBOX_EIP_ALLOCATION_ID: eipalloc-1796f61b
```
dm1tr1yvovk: look like it could be repository level variable

dm1tr1yvovk:
Stepping back from the implementation details: I think the cross-repo EIP mutex part of this PR is solving an artificially scarce resource with too much machinery, and we should rescope.

**The core observation:** the entire `eip-mutex.sh` apparatus (acquire / wait-until-free / verify / release / reap-if-stopped), the cross-repo steal hazard with `mcp-server-namecheap`, and part of the IAM ask exist for exactly one reason — two repos share one whitelisted IP. But that scarcity is artificial: Namecheap API whitelisting allows up to 20 IPs, and an extra Elastic IP costs ~$4/month. And as the PR description itself acknowledges, after the boot-timing fix the mutex degraded to *detect-after-the-fact* rather than *prevent* — so even with all this complexity, collisions with `mcp-server-namecheap` remain possible in both repos.

**Proposal:** allocate a dedicated Elastic IP per repo and whitelist both on the Namecheap side.

- `mcp-server-namecheap` gets its own EIP; each repo keeps its allocation ID in **repo variables** (e.g. `vars.SANDBOX_EIP_ALLOCATION_ID`) instead of hardcoding it in the workflow.
- With no contention, the EIP just stays attached to the warm-pool instance permanently: attached at cold launch via the action's existing `eip-allocation-id` input, survives stop/start, auto-disassociated by the nightly terminate, re-associated next morning. No timing dance, and the no-NAT-gateway boot constraint is naturally satisfied.
- `eip-mutex.sh` gets deleted entirely; `ec2:DisassociateAddress` drops out of the IAM ask (the warm-pool grants — `StopInstances`/`StartInstances`/`ModifyInstanceAttribute`, plus `GetConsoleOutput` for the boot diagnostics this PR discovered were silently broken — are still needed and worth granting regardless).

**What stays — the genuinely valuable parts of this PR are orthogonal to the IP question and should be kept:**

- warm-pool lifecycle (`reuse: stop` + `reuse-pool-tag`) — the actual speedup this was all for;
- `hygiene-sweep.sh` pre/post-job disk wipes;
- `cleanup-ec2-runners.yml` nightly drain + leak reaper;
- the one-line "assert public IP == whitelisted EIP before `make testacc`" sanity check (cheap, catches misconfiguration).

One caveat to check while we're at it: separate IPs fix IP contention, but if both repos' acceptance suites hit the *same sandbox account/domain*, concurrent runs can still race on the DNS records themselves. If that's the case, `mcp-server-namecheap` should get its own sandbox domain at the same time it gets its own IP.

Net effect: the riskiest, least-provable part of this PR (best-effort cross-repo locking) becomes a $4/month line item, the remaining diff is much easier to review, and the IAM ask shrinks.

dm1tr1yvovk (commented):
Direction looks right after the mutex retirement — warm pool, hygiene sweeps, and the nightly drain are exactly the parts worth keeping. Five inline comments below: two I'd consider blocking (the `vars.` switch and the untagged action pin), three doc-accuracy fixes that cost minutes now and prevent real confusion later.

One more that can't take an inline anchor because the PR doesn't touch those lines: the workflow-queue step is still named "Wait for older acceptance runs to release the EIP" and its comment still says the runner is *ephemeral* and that "stop-runner … terminates the host". Under this PR the EIP is never released per-run and stop-runner stops rather than terminates — the host only dies in the nightly drain. Please update the step name/comment to match the new lifecycle.

.github/workflows/ci.yml
dm1tr1yvovk: Two things here:

1. As proposed in the PR thread / #279: please source this from a repo variable rather than a literal — `SANDBOX_EIP_ALLOCATION_ID: ${{ vars.SANDBOX_EIP_ALLOCATION_ID }}`. The workflow already uses this pattern elsewhere (`vars.APP_CLIENT_ID` in cleanup-ec2-runners.yml), and it's what lets other repos adopt the same workflow with their own dedicated EIP without editing YAML.
2. The comment is factually wrong as written: `cleanup-ec2-runners.yml` has **no** top-level `env:` block and never declares this variable — the allocation ID appears there only in a prose header comment. "Keep the two in sync" points at something that doesn't exist. Switching to `vars.` makes the whole sync problem (and this comment) disappear.

.github/workflows/ci.yml
```diff
@@ -426,7 +433,7 @@ jobs:
           aws-region: ${{ secrets.AWS_REGION }}
       - name: Start EC2 runner
         id: start-ec2-runner
-        uses: namecheap/ec2-github-runner@c5db9f03dff1e63a46d40c9b94a49fc5ee6feff1 # v4.0.0 @ 2026-07-02
+        uses: namecheap/ec2-github-runner@65fbe4f0c1181c60d4df2bd3b33b8893e21fbb43 # post-v4.0.0 main (2026-07-06): reuse:stop warm-pool registration fixes (#62/#64/#65/#66)
```
dm1tr1yvovk: This pins to an **untagged main-branch commit**, and the trailing comment (`# post-v4.0.0 main (2026-07-06): …`) doesn't conform to the `# v<semver>` format SECURITY_COMPLIANCE.md requires for every third-party action pin (all prior rotations used `# v4.0.0 @ 2026-07-02` style). Totally understandable while the #62/#64/#65/#66 fixes were being landed live, but before merge: cut a `v4.0.1` tag in `namecheap/ec2-github-runner` and re-pin here (and the identical pin in the stop-runner job) to the tagged SHA with a conforming comment.

.github/workflows/cleanup-ec2-runners.yml
```diff
@@ -0,0 +1,135 @@
...
+      # ANDed with the matching pass kind; every placeholder step negates only
+      # the schedule/dry_run half, keeping the same kind check, so exactly one
+      # of the two ever fires for a given pass.
+
+      - name: Nightly full drain
+        if: ${{ (github.event_name == 'schedule' || inputs.dry_run == false) && steps.pass.outputs.kind == 'nightly-drain' }}
+        # reaper-stopped-max-age is deliberately tiny (1) so the day's pool
+        # instance -- stopped anywhere from seconds to ~24h earlier -- is
+        # always older than the threshold and gets terminated.
+        uses: namecheap/ec2-github-runner@453ef2e080751b77ea3abd1a6fbb0cc27832fbdc # v4.0.0 @ 2026-07-06: includes #62 phone-home fix + #64 GH_REPO_URL sed-delimiter fix + #65 phone-home error detail
```
dm1tr1yvovk: This pins `ec2-github-runner` at `453ef2e0` while ci.yml uses `65fbe4f0` — two versions of the same action managing the same warm pool, and this older one is missing the #66 warm-start user-data fix. SECURITY_COMPLIANCE.md's existing rule keeps this action "at the same SHA" across jobs, and the split isn't documented anywhere. The `v4.0.1` tag suggested on the ci.yml pin resolves this too: pin all four `uses:` sites (start, stop, and both cleanup steps here) to the same tagged SHA.

SECURITY_COMPLIANCE.md
```diff
@@ -135,12 +138,99 @@ self-hosted runner action:
...
+  (`eipalloc-1796f61b`) is passed to the action's `eip-allocation-id` input,
+  which associates it on every cold launch — that is what gives the instance
+  connectivity during its own bootstrap (the fix for the CloudTrail-confirmed
+  cold-boot failures) and is also the IP the Namecheap sandbox API allows. The
+  EIP is deliberately left attached to the pool instance across every warm
+  stop/start cycle rather than released per stop, because the action's
+  warm-restart path never re-associates it. **This repository is the sole
+  user of this allocation** (`mcp-server-namecheap` moved to its own dedicated
+  EIP — see the "Dedicated per-repo sandbox EIP" milestone, #282 /
+  `mcp-server-namecheap#16`), so there is no cross-repo contention and no lock
```
dm1tr1yvovk: This overstates where things actually are: `mcp-server-namecheap` has **not** moved to its own dedicated EIP — per the #282 thread, its Acceptance (EC2) workflow was *disabled* as a stopgap, and both #282 and mcp-server-namecheap#16 are still open. As written, the "sole user" guarantee sounds architectural when it currently rests on a disabled workflow that anyone with write access could re-enable. Since this doc is the authoritative compliance reference, it should say that honestly, e.g.: "sole user of this allocation (mcp-server-namecheap's acceptance workflow is disabled pending its own dedicated EIP — #282 / mcp-server-namecheap#16)". Same wording issue in the "One EIP per repo" bullet further down ("each repo … now owns its own dedicated whitelisted EIP" — not yet true).

SECURITY_COMPLIANCE.md
```diff
@@ -135,12 +138,99 @@ self-hosted runner action:
...
+  terminating the pool instance, which is normally the nightly full drain but
+  can occasionally be the leak-reaper pass instead, if the pool instance
+  happens to sit stopped past its default `reaper-stopped-max-age` before the
+  next drain runs (see below).
+- **IAM prerequisite.** The CI AWS identity requires `ec2:AssociateAddress`
+  and `ec2:DescribeInstances` (used by the action to associate the EIP and
+  manage the instance) plus `ec2:DescribeAddresses` (to resolve the
+  whitelisted EIP's public IP for the acceptance gate). `ec2:DisassociateAddress`
+  is **not** required — the former cross-repo EIP reaper that used it has been
+  removed along with the shared-EIP mutex.
```
dm1tr1yvovk: This IAM list is incomplete against what this PR's own investigation established: `reuse: stop` needs `ec2:StopInstances`, `ec2:StartInstances`, and `ec2:ModifyInstanceAttribute` (warm-stop + user-data rewrite on restart), and the bootstrap fast-fail/console diagnostics need `ec2:GetConsoleOutput` + `ec2:DescribeTags` — the exact gaps enumerated in the PR thread's policy-diff comment. An AWS admin following this bullet as written would grant a policy under which warm-pool cycles fail with `UnauthorizedOperation`. Please fold the full list from that comment into this bullet so the doc matches reality.

.github/workflows/ci.yml
kurok: Done in 09a5167 — added repo variable `SANDBOX_EIP_ALLOCATION_ID=eipalloc-1796f61b` and the env now reads `${{ vars.SANDBOX_EIP_ALLOCATION_ID }}`.

kurok: Done in 09a5167. The env now sources `vars.SANDBOX_EIP_ALLOCATION_ID` (variable set to `eipalloc-1796f61b`), so another repo can adopt this workflow with its own dedicated EIP without editing YAML. You're also right about (2): the "cleanup-ec2-runners.yml repeats this literal / keep in sync" claim was false — cleanup only mentions the allocation in a prose comment — so that comment is gone.

.github/workflows/ci.yml
kurok: Done — cut `v4.0.1` at 65fbe4f in namecheap/ec2-github-runner (https://github.com/namecheap/ec2-github-runner/releases/tag/v4.0.1) and re-pinned both ci.yml sites (start + stop) to that SHA under a conforming `# v4.0.1 @ 2026-07-08` comment in 09a5167. The `v4` major tag was intentionally left unmoved.

.github/workflows/cleanup-ec2-runners.yml
kurok: Fixed in 09a5167 — both cleanup steps were on 453ef2e (v4.0.0, missing the #66 warm-start fix); they now pin the same `65fbe4f` `# v4.0.1` as ci.yml. All four `uses:` sites (ci start+stop, cleanup ×2) are identical.

SECURITY_COMPLIANCE.md
kurok: Reworded in 09a5167. The dedicated-EIP bullet now says this repo is *currently* the sole user **because mcp-server-namecheap's Acceptance (EC2) workflow is disabled as an interim stopgap** (#282 + mcp-server-namecheap#16 still open), explicitly noting it's not yet a permanent per-repo split and that re-enabling that workflow before it has its own EIP would reintroduce the contention. The "One EIP per repo" bullet now says the migration is in progress, not complete.

SECURITY_COMPLIANCE.md
kurok: Expanded in 09a5167 to the full set from your policy-diff comment: instance lifecycle (RunInstances/TerminateInstances/CreateTags/DescribeImages/DescribeInstances/DescribeInstanceStatus), warm-pool reuse:stop (StopInstances/StartInstances/ModifyInstanceAttribute), EIP (AssociateAddress + DescribeAddresses — `DisassociateAddress` dropped now the mutex is gone), and bootstrap diagnostics (GetConsoleOutput + DescribeTags).

dm1tr1yvovk (dismissed):
All review items verified fixed: the EIP allocation now comes from the `SANDBOX_EIP_ALLOCATION_ID` repo variable, all four `ec2-github-runner` sites pin the same tagged `v4.0.1` SHA with conforming comments, the compliance doc now states the mcp-server-namecheap stopgap honestly, and the IAM list is complete. The explicit re-association on every start (`c4ede4ca`) is a nice catch — and only safe because the EIP is now dedicated to this repo.

One non-blocking leftover (from the previous review body, easy to miss): the queue step at ci.yml:415 is still named "Wait for older acceptance runs to release the EIP" with a comment describing the old ephemeral/terminate lifecycle. Fine to fix here or in a follow-up.

Approving — thanks for the fast turnaround on all of it.

kurok:
Addressed the non-blocking leftover in c60c7a4: the workflow-queue step is renamed to **"Wait for older acceptance runs to finish (single sandbox EIP)"** and its comment rewritten for the warm-pool model (FIFO serialization so only one run holds the single sandbox EIP / pooled instance at a time; `stop-runner` stops rather than terminates, and the EIP stays attached to the stopped instance). Comment-only, no behavior change.

kurok:
@dm1tr1yvovk — ready for another look. All of your feedback is addressed and CI is green (warm reuse working end-to-end; last several runs pass).

**Your 6 inline comments** (all in `09a5167`, replies on each thread):
- EIP allocation now sourced from `vars.SANDBOX_EIP_ALLOCATION_ID` (variable set); false env comment removed.
- Cut `ec2-github-runner` **v4.0.1** at `65fbe4f` and pinned all four `uses:` sites to it with a conforming `# v4.0.1` comment (cleanup was on the stale `453ef2e`).
- `SECURITY_COMPLIANCE.md`: dropped the overstated "mcp moved to its own EIP" claim (it's *disabled as an interim stopgap*; #282 + mcp-server-namecheap#16 still open) and completed the IAM list (warm-pool `StopInstances`/`StartInstances`/`ModifyInstanceAttribute` + diagnostics `GetConsoleOutput`/`DescribeTags`).

**Non-blocking nit** (`c60c7a4`): queue step renamed/re-commented for the warm-pool model.

**Two CI issues surfaced while verifying** (worth a glance):
- The whitelisted EIP had drifted off the long-lived pool instance (the scheduled reaper isn't active until this merges to `master`). Fixed by a **"Pre-attach sandbox EIP"** step (`ad7c2c3`) that attaches it to the stopped pool instance *before* `mode: start`, so the runner boots on the whitelisted IP — associating *after* start changed a registered runner's IP and hung the job, so that approach was reverted.

Thanks for the thorough review.

dm1tr1yvovk (approved):
Re-approving at `c60c7a4e` — all review items are now closed, including the last leftover (the queue step rename/re-comment for the warm-pool model). Summary of the final state:

**Review items — all resolved:**
- EIP allocation ID sourced from the `SANDBOX_EIP_ALLOCATION_ID` repo variable (no hardcoded literal, false "keep in sync" comment gone).
- All four `ec2-github-runner` `uses:` sites pin the same tagged SHA (`65fbe4f` / `# v4.0.1 @ 2026-07-08`), conforming to the SECURITY_COMPLIANCE.md pin convention.
- SECURITY_COMPLIANCE.md now states the mcp-server-namecheap situation accurately (workflow disabled as an interim stopgap; #282 / mcp-server-namecheap#16 still open) and carries the complete IAM list including the warm-pool and diagnostics grants.
- Queue step name/comment updated for the warm-pool lifecycle.

**Live verification (run 28941899602, push @ 12:16 UTC):** the start-runner log shows `Pre-attached eipalloc-1796f61b to stopped warm-pool instance i-06fa6607953cc353d` followed by `Reused warm-pool EC2 instance i-06fa6607953cc353d`; runner start took ~69s (vs the former 2–4 min cold boot) and the real sandbox acceptance suite passed in ~90s. Three successful real acceptance runs today across warm stop/start cycles — the new lifecycle is demonstrably the code path in use, not just green checks.

For anyone reading later re "Acceptance test: skipped": that's the `pull_request`-event twin run, where the real acceptance jobs are intentionally gated (`if: github.event_name == 'push'`, the fork-safety guard that is load-bearing for warm-pool disk reuse). The `push`-event run of the same commit executes the real suite; the PR run covers the mock matrix.

Good to merge.

PR #284, opened 2026-07-09
feat: add namecheap_nameserver resource (register personal nameservers)

### LINKED ISSUE #76: Feature request: `namecheap_nameserver` resource to register new nameservers
I'd like to also manage custom nameservers w/ terraform, i.e. use the following API: https://www.namecheap.com/support/api/methods/domains-ns/create/.

However, this provider does not yet support this. Are there any plans to add this?

kurok:
## What

Adds a new `namecheap_nameserver` resource that registers and manages **personal (glue/vanity) nameservers** — such as `ns1.example.com` pointed at a glue IP — for a domain on the account.

## Why

Requested in #76. The provider previously could only *assign* custom nameservers to a domain (`namecheap_domain_records.nameservers`, which calls `domains.dns.setCustom`). It had no way to *register* a personal nameserver host + glue IP. This resource fills that gap by wrapping the `domains.ns.*` API family.

## SDK methods used

The pinned `go-namecheap-sdk/v2 v2.7.0` already exposes per-record atomic operations, which map cleanly onto CRUD:

| CRUD | SDK call |
|------|----------|
| Create | `client.DomainsNS.CreateWithContext(ctx, sld, tld, nameserver, ip)` |
| Read | `client.DomainsNS.GetInfoWithContext(ctx, sld, tld, nameserver)` |
| Update | `client.DomainsNS.UpdateWithContext(ctx, sld, tld, nameserver, oldIP, ip)` |
| Delete | `client.DomainsNS.DeleteWithContext(ctx, sld, tld, nameserver)` |

`domain` is parsed into SLD/TLD via `namecheap.ParseDomain`. API errors are surfaced through the existing `diagFromClientError` mapper.

## Schema

- `domain` (Required, ForceNew) — root domain, validated with the existing `validateDomainIsNotSubdomain`.
- `nameserver` (Required, ForceNew) — FQDN of the nameserver host.
- `ip` (Required) — glue IP; changeable in place (triggers `domains.ns.update`).

## ID / import scheme

Composite ID `<domain>/<nameserver>` (neither a domain nor a nameserver hostname contains a slash, so it parses unambiguously):

```
terraform import namecheap_nameserver.ns1 example.com/ns1.example.com
```

## Testing performed

- `gofmt -l namecheap/` clean; `go build ./...`; `go vet ./...` (and `-tags testacc`) clean; `golangci-lint run` → 0 issues; `make build` OK.
- Unit tests (`go test ./namecheap/...`, also `-race`): create / read / read-not-found (drops from state) / update / delete / API-error / import (valid + invalid IDs) / ID composition / domain split.
- Stateful mock-server acceptance lifecycle (`make testacc-mock`, real terraform binary): create → update (IP change) → import (`ImportStateVerify`) → destroy, asserting both Terraform state and mock backend. Extends the existing mock to serve the `domains.ns.*` commands.
- Live-sandbox acceptance suite was **not** run (no credentials); no live tests were added.

Adds Terraform Registry docs at `docs/resources/nameserver.md`.

Note: `provider.go`'s `ResourcesMap` is edited to register the resource (sibling PRs may also touch this file).

Closes #76

### DIFF
--- a/namecheap/mock_personal_nameserver_test.go
+++ b/namecheap/mock_personal_nameserver_test.go
@@ -94,3 +94,48 @@ resource "namecheap_personal_nameserver" "test" {
 		},
 	})
 }
+
+// TestAccMockPersonalNameserverPair covers the headline use case — registering a
+// pair of personal (vanity) nameservers (ns1/ns2) under a single domain so the
+// domain can run on its own nameservers. It asserts both hosts are persisted
+// with their glue IPs and both are removed on destroy, against the stateful
+// mock.
+func TestAccMockPersonalNameserverPair(t *testing.T) {
+	m := newNamecheapMock(t)
+	const domain = "mock-example.com"
+	const ns1 = "ns1.mock-example.com"
+	const ns2 = "ns2.mock-example.com"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { mockPreCheck(t, m) },
+		ProviderFactories: mockProviderFactories(),
+		CheckDestroy: resource.ComposeTestCheckFunc(
+			mockCheckNameserverAbsent(m, domain, ns1),
+			mockCheckNameserverAbsent(m, domain, ns2),
+		),
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(`
+resource "namecheap_personal_nameserver" "ns1" {
+  domain     = %[1]q
+  nameserver = %[2]q
+  ip         = "192.0.2.1"
+}
+
+resource "namecheap_personal_nameserver" "ns2" {
+  domain     = %[1]q
+  nameserver = %[3]q
+  ip         = "192.0.2.2"
+}
+`, domain, ns1, ns2),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("namecheap_personal_nameserver.ns1", "ip", "192.0.2.1"),
+					resource.TestCheckResourceAttr("namecheap_personal_nameserver.ns1", "id", domain+"/"+ns1),
+					resource.TestCheckResourceAttr("namecheap_personal_nameserver.ns2", "ip", "192.0.2.2"),
+					mockCheckNameserverIP(m, domain, ns1, "192.0.2.1"),
+					mockCheckNameserverIP(m, domain, ns2, "192.0.2.2"),
+				),
+			},
+		},
+	})
+}
--- a/namecheap/mock_personal_nameserver_test.go
+++ b/namecheap/mock_personal_nameserver_test.go
@@ -45,13 +45,13 @@ func mockCheckNameserverAbsent(m *namecheapMock, domain, nameserver string) reso
 }
 
 // TestAccMockNameserverLifecycle drives a full create -> update (IP change) ->
-// import -> destroy lifecycle for namecheap_nameserver against the stateful
-// mock, asserting both Terraform state and the mock backend at each step.
+// import -> destroy lifecycle for namecheap_personal_nameserver against the
+// stateful mock, asserting both Terraform state and the mock backend at each step.
 func TestAccMockNameserverLifecycle(t *testing.T) {
 	m := newNamecheapMock(t)
 	const domain = "mock-example.com"
 	const nameserver = "ns1.mock-example.com"
-	const resourceName = "namecheap_nameserver.test"
+	const resourceName = "namecheap_personal_nameserver.test"
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:          func() { mockPreCheck(t, m) },
@@ -60,7 +60,7 @@ func TestAccMockNameserverLifecycle(t *testing.T) {
 		Steps: []resource.TestStep{
 			{
 				Config: fmt.Sprintf(`
-resource "namecheap_nameserver" "test" {
+resource "namecheap_personal_nameserver" "test" {
   domain     = "%s"
   nameserver = "%s"
   ip         = "1.2.3.4"
@@ -74,7 +74,7 @@ resource "namecheap_nameserver" "test" {
 			},
 			{
 				Config: fmt.Sprintf(`
-resource "namecheap_nameserver" "test" {
+resource "namecheap_personal_nameserver" "test" {
   domain     = "%s"
   nameserver = "%s"
   ip         = "5.6.7.8"
--- a/namecheap/provider.go
+++ b/namecheap/provider.go
@@ -104,8 +104,8 @@ func Provider() *schema.Provider {
 			},
 		},
 		ResourcesMap: map[string]*schema.Resource{
-			"namecheap_domain_records": resourceNamecheapDomainRecords(),
-			"namecheap_nameserver":     resourceNamecheapNameserver(),
+			"namecheap_domain_records":      resourceNamecheapDomainRecords(),
+			"namecheap_personal_nameserver": resourceNamecheapPersonalNameserver(),
 		},
 		ConfigureContextFunc: configureContext,
 	}
--- a/namecheap/provider.go
+++ b/namecheap/provider.go
@@ -105,6 +105,7 @@ func Provider() *schema.Provider {
 		},
 		ResourcesMap: map[string]*schema.Resource{
 			"namecheap_domain_records": resourceNamecheapDomainRecords(),
+			"namecheap_nameserver":     resourceNamecheapNameserver(),
 		},
 		ConfigureContextFunc: configureContext,
 	}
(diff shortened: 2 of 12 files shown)


namecheap/provider.go
```diff
@@ -105,6 +105,7 @@ func Provider() *schema.Provider {
 		},
 		ResourcesMap: map[string]*schema.Resource{
 			"namecheap_domain_records": resourceNamecheapDomainRecords(),
+			"namecheap_nameserver":     resourceNamecheapNameserver(),
```
dm1tr1yvovk: maybe better to name it as namecheap_personal_nameserver to indicate that it is about glue records configuration etc, not just nameservers

namecheap/mock_personal_nameserver_test.go
```diff
@@ -0,0 +1,96 @@
+//go:build testacc
```
dm1tr1yvovk, line 1: looks like extra commnet

dm1tr1yvovk (commented):
can you create real acceptance tests to reach api to verify how this works?

dm1tr1yvovk (changes requested):

namecheap/provider.go
kurok: Done in 28063f2 — renamed the resource type, its registration function (`resourceNamecheapPersonalNameserver`), the source/test/mock files and the registry doc page to `namecheap_personal_nameserver`. `make build/vet/lint/test/testacc-mock` all green.

namecheap/mock_personal_nameserver_test.go
kurok: This line is the Go build constraint `//go:build testacc`, not a stray comment — it gates the file to the `testacc` build tag, matching every other `mock_*_test.go` in the package (`mock_server_test.go`, `mock_acceptance_test.go`, `mock_data_source_test.go`). The file references testacc-only symbols (`namecheapMock`, `newNamecheapMock`, `mockProviderFactories`, `resource.Test`), so without the tag the default `go build`/`go vet` would fail to compile. Leaving it in place to match the established pattern — happy to reorganize the mock helpers if you'd prefer a different split.

dm1tr1yvovk:
### Code review

Found 1 issue:

1. `resourceNameserverRead` never takes its "drop from state" path when a personal nameserver is deleted outside Terraform. The real API — and this PR's own stateful mock (`apiErrorXML("5013160", "Nameserver not found")`) — signals a missing nameserver as a Status=ERROR response, which the SDK surfaces as a non-nil error from `GetInfoWithContext`. That hits the `if err != nil` branch and returns a hard diagnostic (`diagFromClientError` has no case for 5013160, so it falls through to a generic error), so `terraform plan`/`refresh` fails instead of planning a recreate. The `resp == nil || resp.DomainNameserverInfoResult == nil` branch below it (with the comment promising exactly this behavior) is unreachable for the real not-found path — the SDK only yields a nil-result-with-nil-error for a Status=OK empty body, which the API doesn't produce here. The unit test `TestResourceNameserverReadNotFound` passes only because its fixture uses that fabricated OK-with-empty-body shape rather than the error response the mock server itself returns for the same scenario. Suggested fix: detect `*namecheap.APIError` with `Number == 5013160` in Read (via `errors.As`), call `data.SetId("")` and return nil; update the unit test fixture to use the API-error shape. Worth applying the same treatment to Delete for idempotency.

https://github.com/namecheap/terraform-provider-namecheap/blob/e32637b96dbc89c5b39f0695f92192196307e573/namecheap/namecheap_personal_nameserver.go#L107-L118

Contradicting mock behavior for the same scenario:

https://github.com/namecheap/terraform-provider-namecheap/blob/e32637b96dbc89c5b39f0695f92192196307e573/namecheap/mock_server_test.go#L166-L168

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

kurok:
Thanks @dm1tr1yvovk — both points addressed:

**1. Real acceptance tests (`can you create real acceptance tests to reach api...`)** — added in `8a89b0d`:
- `TestAccNamecheapPersonalNameserver` (live sandbox, `TF_ACC`-gated): registers an ns1/ns2 vanity pair under the test domain, changes ns1's glue IP in place, imports by `<domain>/<nameserver>`, and destroys — asserting Terraform state **and** the live API via `domains.ns.getInfo` at each step, plus a `CheckDestroy` that both hosts are deregistered.
- `TestAccMockPersonalNameserverPair` for the same use case against the stateful mock.
- Combined with `e32637b` (CI now runs the live sandbox suite on same-repo PRs), this exercises the resource against the real API on the PR itself.

**2. Not-found drift bug** (`resourceNameserverRead` never dropping from state) — fixed in `b217785`. You were exactly right: the API returns `Status=ERROR` "Nameserver not found" (5013160), the SDK surfaces it as a non-nil `*namecheap.APIError`, and `diagFromClientError` had no 5013160 case, so Read failed the refresh instead of planning a recreate — the `resp == nil` branch was unreachable for the real shape. Now Read detects the 5013160 `APIError` (via `errors.As`), clears the ID, and returns nil; Delete swallows the same code for idempotent destroy. The not-found unit test now uses the real `Status=ERROR` fixture (matching what the stateful mock returns), with added tests for the empty-result defensive path, a non-not-found hard error (still surfaced), and idempotent delete.

`make build/format/check/lint/test/testacc-mock` all green; project coverage 91.4%.

dm1tr1yvovk (dismissed):
### Review summary

Approving. The resource implementation, the follow-up fix, and the test coverage were all verified:

**Code review** ([earlier comment](https://github.com/namecheap/terraform-provider-namecheap/pull/284#issuecomment-4925044633)): one real bug was found — `Read` surfaced the API's `Status=ERROR` "Nameserver not found" (5013160) as a hard diagnostic instead of dropping the resource from state, so an out-of-band deletion broke `terraform plan`/`refresh`. Commit b217785 fixes it exactly right: `errors.As`-based detection of code 5013160 in `Read` (SetId("") → recreate plan) and in `Delete` (idempotent destroy), with the old nil-response branch kept as a defensive path.

**Tests**: the previously-misleading `TestResourceNameserverReadNotFound` fixture now uses the real API-error shape (matching the stateful mock), plus new tests for the defensive empty-result path, non-not-found API errors still surfacing as hard errors, and idempotent delete. Live sandbox coverage (`TestAccNamecheapPersonalNameserver`) delivers the full vanity-NS-pair lifecycle — create, in-place glue IP update, import with `ImportStateVerify` — asserted against `domains.ns.getInfo` at every step, with collision-proof per-run naming.

**Verified locally at head 3b3c9fc**: `go build`, `go vet`, full unit suite, and both mock acceptance tests (`TestAccMockNameserverLifecycle`, `TestAccMockPersonalNameserverPair`) pass. All commits carry DCO sign-offs; Check, CodeQL, Trivy, govulncheck are green.

Note: the pending "Acceptance test" job on the latest run is stuck in queue due to a warm-pool runner-registration flake in ec2-github-runner (the started instance never re-registered under the new label), not due to anything in this PR — being handled separately. The live suite passed on this branch at run 29017428944.

dm1tr1yvovk:
The stuck "Acceptance test" job on run [29019773728](https://github.com/namecheap/terraform-provider-namecheap/actions/runs/29019773728) is now tracked in namecheap/ec2-github-runner#67 — a `reuse: stop` warm-restart race where `mode: start` reports the runner online but no runner ever serves the label, so the job queues indefinitely and `stop-runner` never fires. Not caused by this PR; the live suite passed on this branch at run [29017428944](https://github.com/namecheap/terraform-provider-namecheap/actions/runs/29017428944). Unblocking: cancel the stuck run (fires `stop-runner`, stops the instance), then re-run.

kurok:
@dm1tr1yvovk this is ready for another look — all review points addressed and CI is green (19 passing, live sandbox acceptance included).

Summary of what changed since your review:

- **Rename** → `namecheap_personal_nameserver` (type, registration func, source/test/mock files, registry doc page). `28063f2`
- **Not-found drift bug** (your code review): Read now detects the `5013160` "Nameserver not found" `APIError` and drops the resource from state (plans a recreate) instead of hard-failing the refresh; Delete swallows the same code for idempotent destroy. Unit tests switched to the real `Status=ERROR` fixture, with added coverage for the empty-result path, a still-surfaced non-not-found error, and idempotent delete. `b217785`
- **Real acceptance tests reaching the API** (your request): `TestAccNamecheapPersonalNameserver` registers an ns1/ns2 vanity pair on the sandbox domain, changes a glue IP in place, imports, and destroys — asserting state and the live API (`domains.ns.getInfo`) at each step; plus `TestAccMockPersonalNameserverPair`. `8a89b0d`
- Fixed the glue IPs after the live run surfaced a real API constraint: Namecheap rejects reserved ranges for glue records (`IP address ... is reserved (3024278)`), so the test and the doc example now use routable public IPs, and the doc has a note listing the reserved ranges. `8516fae`

The live acceptance suite now runs on same-repo PRs (per the CI change in this PR), so it exercised all of the above against the real sandbox on this PR.

dm1tr1yvovk (approved):

PR #285, opened 2026-07-09
feat: add namecheap_domain_contacts resource (WHOIS contact management)

### LINKED ISSUE #74: Ability to update the domain contact record information
You should be able to update the  contact information for a domain using the terraform provider. Currently only `domain_records` can be managed, not a `domain`

### Terraform Version

Terraform v1.2.4
on linux_amd64

### Namecheap provider version

+ provider registry.terraform.io/namecheap/namecheap v2.1.0

### Affected Resource(s)

* namecheap_domain

### LINKED ISSUE #252: New resource: namecheap_domain_contacts — WHOIS contact management (#74)
## Problem

Open issue #74: users cannot update domain WHOIS contact information via the provider. Contact data is compliance-relevant (ICANN accuracy obligations, GDPR-driven privacy setups) and changes on real events (office moves, org renames, role-account rotations) — exactly the kind of fleet-wide change (50 domains × same address update) that begs for `for_each` instead of 50 dashboard sessions. Unblocks with go-namecheap-sdk#114 (`getContacts`/`setContacts` + shared `ContactInfo`).

### `namecheap_domain_contacts`

```hcl
resource "namecheap_domain_contacts" "main" {
  domain = "example.com"

  registrant {
    first_name    = "Jane"
    last_name     = "Doe"
    organization  = "Example Corp"
    address1      = "..."
    city          = "..."
    country       = "PT"
    phone         = "+351.123456789"
    email_address = "hostmaster@example.com"
    # full field set per docs/namecheap-api-v2.md contact tables
  }
  tech        { ... }   # optional; default = registrant
  admin       { ... }   # optional; default = registrant
  aux_billing { ... }   # optional; default = registrant
}
```

- Standalone resource for domains whose *registration* isn't Terraform-managed (imported portfolios — the dominant case today). Mutually exclusive with inline `contacts` on `namecheap_domain` for the same domain (documented, same pattern as the records resources).
- **Default-to-registrant** semantics for omitted blocks are explicit and shown in the plan (no invisible API-side defaulting).
- Plan-time validation of the full per-country required-field matrix where the API documents it (phone format `+NNN.NNNNNNN`, country codes, required address fields) — contact rejections are slow server-side; catch them at plan.
- **Privacy interplay documented**: with WHOIS privacy enabled, registry-visible data is the privacy service's — this resource manages the *underlying* registrant data; the docs section spells out what changes where (the support question this resource will otherwise generate).
- Import: `terraform import namecheap_domain_contacts.main example.com` seeds state from `getContacts`.

## Acceptance criteria

- [ ] Lifecycle matrix against the mock: create(set)/read/update per contact block/import/destroy (destroy = state-remove with warning — contacts can't be "deleted", documented like the domain resource's abandon semantics).
- [ ] Field mapping is exhaustive against the SDK `ContactInfo` struct — a completeness test enumerates struct fields vs schema attributes so SDK additions fail loudly here instead of silently dropping data.
- [ ] Omitted tech/admin/aux blocks apply registrant values and show that in plan (test asserts both the API payload and the plan rendering).
- [ ] Validators: phone format, ISO country codes, email syntax, per-block required fields — table-driven plan-time tests; a fixture-based test covers one country with extended requirements.
- [ ] Drift: dashboard-side contact change surfaces on refresh (mock-scripted).
- [ ] Mutual exclusion with `namecheap_domain.contacts` documented in both resources' registry docs; fleet-update example (`for_each` over 3 domains) in examples/.
- [ ] Issue #74 linked (`Closes #74`); sandbox suite exercises set+get round-trip on the test domain.
- [ ] `make build format check lint test testacc` green; conventional commits; DCO.

## Out of scope

- Address-book (`users.address`) backed contact references (nice follow-up once go-namecheap-sdk#117 lands: `contacts_from_address_book = "default"`; noted in the epic, not blocking).
- WHOIS privacy toggling (domain resource / SDK #118).

Depends on: go-namecheap-sdk#114, framework migration (B), testing infrastructure. Part of the provider 3.0 roadmap (see tracking issue).

kurok:
## What

Adds a new resource **`namecheap_domain_contacts`** for managing a domain's WHOIS contacts — the `registrant`, `tech`, `admin` and `aux_billing` blocks — as a standalone resource (well suited to imported portfolios and fleet-wide `for_each` updates).

## Why

Closes the long-standing request to manage domain contact information from Terraform. Contact data is compliance-relevant (ICANN accuracy, GDPR privacy setups) and changes on real events (office moves, org renames, role-account rotation) — the kind of fleet change that begs for `for_each` instead of dashboard clicking.

## SDK methods used

Built on the already-pinned `go-namecheap-sdk/v2 v2.7.0` (no SDK bump):

- `client.Domains.SetContactsWithContext` — Create and Update.
- `client.Domains.GetContactsWithContext` — Read and Import.
- The shared `namecheap.ContactInfo` struct — schema modeled 1:1 on its fields.

## Behavior

- **Create/Update** → one `setContacts` call. **Read** → `getContacts` mapped back into state, so a dashboard-side change surfaces as drift on refresh.
- **Registrant required**; `tech`/`admin`/`aux_billing` optional and **default to the registrant** when omitted. `setContacts` requires all four blocks, so the provider fills the omitted ones — and does so in `CustomizeDiff`, so the resolved values are **shown in the plan** rather than applied invisibly server-side.
- **Delete = state-only removal.** The Namecheap API has no delete-contacts operation, so destroying the resource stops Terraform managing the contacts but leaves the last-applied values on the domain. This is surfaced as a **warning diagnostic** (abandon-, not delete-, semantics). See caveat below.
- **Import** by domain name (`terraform import namecheap_domain_contacts.main example.com`) seeds state from `getContacts`.
- **Plan-time validators** for `phone` (`+NNN.NNNNNNN`), `country` (ISO alpha-2) and `email_address` format; per-block required fields enforced by the schema.

## Reviewer caveat — Delete semantics

`Delete` intentionally makes **no API call** and only removes the resource from state (with a warning). Namecheap exposes no way to clear contacts, so they persist on the domain at their last-set values. This is documented in the resource doc and in code comments.

## Testing performed

- `make build`, `gofmt -l`, `go vet ./...` (both default and `testacc` tags), `golangci-lint run` — all clean (0 issues).
- **Unit tests**: expand/flatten round-trip, `contactOrDefault`, format validators (table-driven), required/optional field mapping, schema `InternalValidate`, and a **completeness test** that reflects over `ContactInfo` and fails loudly if the SDK adds a field the schema doesn't map.
- **Mock acceptance tests** (`make testacc-mock`, in-process stateful mock, no credentials): full create→update→import→destroy lifecycle; default-to-registrant asserting **both the API payload and the plan-rendered state** plus a stable re-plan; out-of-band drift detection; and import of pre-seeded contacts verified against the imported state.
- **Live-sandbox test** (`TestAccNamecheapDomainContacts`) exercises the set+get round-trip and import on `NAMECHEAP_TEST_DOMAIN`; it **skips without credentials** and was not run against the live API here.

## Docs

`docs/resources/domain_contacts.md` covers usage (including a `for_each` fleet example), the full field reference, default-to-registrant, WHOIS-privacy interplay, mutual-exclusion guidance, import, and destroy semantics.

Note: this PR also edits `namecheap/provider.go` (`ResourcesMap`); sibling PRs touch the same file — expected.

Closes #252
Closes #74

Signed-off-by: kurok <22548029+kurok@users.noreply.github.com>

### DIFF
--- a/docs/resources/domain_contacts.md
+++ b/docs/resources/domain_contacts.md
@@ -1,0 +1,151 @@
+---
+page_title: "namecheap_domain_contacts Resource - terraform-provider-namecheap"
+subcategory: ""
+description: |-
+  Manages a domain's WHOIS contact information (registrant, tech, admin and auxiliary billing contacts).
+---
+
+# namecheap_domain_contacts (Resource)
+
+Manages the WHOIS contact information of a domain in your Namecheap account: the
+`registrant`, `tech`, `admin` and `aux_billing` contact blocks. It maps onto the
+Namecheap `namecheap.domains.getContacts` / `namecheap.domains.setContacts` API.
+
+This is a standalone resource, well suited to domains whose registration is not
+otherwise managed by Terraform (imported portfolios), and to fleet-wide contact
+updates via `for_each`.
+
+## Example Usage
+
+```terraform
+resource "namecheap_domain_contacts" "main" {
+  domain = "example.com"
+
+  registrant {
+    first_name     = "Jane"
+    last_name      = "Doe"
+    organization   = "Example Corp"
+    address1       = "1 Main St"
+    city           = "Lisbon"
+    state_province = "Lisboa"
+    postal_code    = "1000-001"
+    country        = "PT"
+    phone          = "+351.123456789"
+    email_address  = "hostmaster@example.com"
+  }
+
+  # tech, admin and aux_billing are optional. When omitted they default to the
+  # registrant values; the defaulting is shown in the plan (it is not hidden
+  # server-side). Provide a block to override a specific contact:
+  #
+  # tech {
+  #   first_name     = "Tim"
+  #   ...
+  # }
+}
+```
+
+### Fleet update (`for_each`)
+
+Apply the same contact to many domains at once:
+
+```terraform
+locals {
+  managed_domains = ["example.com", "example.net", "example.org"]
+}
+
+resource "namecheap_domain_contacts" "fleet" {
+  for_each = toset(local.managed_domains)
+  domain   = each.value
+
+  registrant {
+    first_name     = "Jane"
+    last_name      = "Doe"
+    organization   = "Example Corp"
+    address1       = "1 Main St"
+    city           = "Lisbon"
+    state_province = "Lisboa"
+    postal_code    = "1000-001"
+    country        = "PT"
+    phone          = "+351.123456789"
+    email_address  = "hostmaster@example.com"
+  }
+}
+```
+
+## Argument Reference
+
+- `domain` - (Required, ForceNew) Purchased available domain name on your account. Must be a registered root domain (e.g. `example.com`), not a subdomain.
+- `registrant` - (Required) The registrant contact (see [nested schema](#nested-schema-for-contact-blocks)).
+- `tech` - (Optional) The tech contact. Defaults to `registrant` when omitted.
+- `admin` - (Optional) The admin contact. Defaults to `registrant` when omitted.
+- `aux_billing` - (Optional) The auxiliary billing contact. Defaults to `registrant` when omitted.
+
+### Nested Schema for contact blocks
+
+Each of `registrant`, `tech`, `admin` and `aux_billing` accepts the same fields.
+
+Required:
+
+- `first_name` - (Required) The contact's first name.
+- `last_name` - (Required) The contact's last name.
+- `address1` - (Required) The primary street address.
+- `city` - (Required) The contact's city.
+- `state_province` - (Required) The state or province.
+- `postal_code` - (Required) The postal/ZIP code.
+- `country` - (Required) The two-letter ISO 3166-1 alpha-2 country code (e.g. `US`, `PT`).
+- `phone` - (Required) The phone number in `+NNN.NNNNNNNNNN` format (e.g. `+1.6613102107`).
+- `email_address` - (Required) The contact e-mail address.
+
+Optional:
+
+- `organization` - (Optional) The contact's organization.
+- `job_title` - (Optional) The contact's job title.
+- `address2` - (Optional) The secondary street address.
+
+~> **Plan-time validation.** `phone`, `country` and `email_address` are validated
+at plan time (format only) to catch obvious mistakes before the slower
+server-side rejection. Namecheap enforces additional per-country and
+ICANN-required-field rules server-side; a valid-looking value can still be
+rejected on apply.
+
+## Default-to-registrant
+
+Omitted `tech`, `admin` and `aux_billing` blocks default to the `registrant`
+values. The Namecheap `setContacts` API requires all four contact blocks, so the
+provider fills the omitted ones with the registrant. This defaulting is applied
+during planning, so the resolved values appear in the plan and in state rather
+than being applied invisibly.
+
+## WHOIS privacy interplay
+
+This resource manages the *underlying* registrant data on the domain. When WHOIS
+privacy (Domain Privacy / WhoisGuard) is enabled, the registry-visible contact is
+the privacy service's proxy, not the values managed here — those remain the
+domain's real contact data behind the privacy service. Toggling WHOIS privacy is
+out of scope for this resource.
+
+## Mutual exclusion
+
+Manage a given domain's contacts in exactly one place. Do not point two
+`namecheap_domain_contacts` resources at the same domain, and do not combine this
+resource with any other mechanism that also sets the domain's contacts — the last
+apply wins and the resources will fight on every plan.
+
+## Import
+
+Domain contacts can be imported by domain name, e.g.,
+
+```terraform
+terraform import namecheap_domain_contacts.main example.com
+```
+
+The import seeds state from `getContacts`.
+
+## Destroy semantics
+
+The Namecheap API has no operation to delete a domain's contacts. Destroying this
+resource removes it from Terraform state (and emits a warning) but leaves the
+last-applied contact values on the domain — analogous to abandoning, rather than
+deleting, the managed data. To change the contacts after destroy, manage them
+again with this resource or edit them in the Namecheap dashboard.
--- a/namecheap/mock_domain_contacts_test.go
+++ b/namecheap/mock_domain_contacts_test.go
@@ -192,30 +192,32 @@ func TestAccMockDomainContactsDrift(t *testing.T) {
 	})
 }
 
-// TestAccMockDomainContactsImport seeds pre-existing contacts (a portfolio
-// domain not yet Terraform-managed) and imports them by domain name.
+// TestAccMockDomainContactsImport creates a domain's contacts, then imports the
+// resource by domain name and verifies every block round-trips through
+// getContacts into Terraform state.
 func TestAccMockDomainContactsImport(t *testing.T) {
 	m := newNamecheapMock(t)
 	const resourceName = "namecheap_domain_contacts.test"
-
-	m.seedContacts(mockContactsDomain, map[string]map[string]string{
-		"Registrant": fullContactWire(),
-		"Tech":       fullContactWire(),
-		"Admin":      fullContactWire(),
-		"AuxBilling": fullContactWire(),
-	})
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:          func() { mockPreCheck(t, m) },
 		ProviderFactories: mockProviderFactories(),
 		Steps: []resource.TestStep{
 			{
-				Config:        registrantOnlyConfig(mockContactsDomain, "jane@example.com"),
+				// Establish the contacts first. The import step below deliberately
+				// carries no Config: an import-only step that also supplies a Config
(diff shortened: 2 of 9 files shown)


dm1tr1yvovk:
# Review summary

Overall the PR is well built: single-source-of-truth field mapping with a reflection-based completeness test, default-to-registrant materialized in `CustomizeDiff` (so it shows in the plan), correct docs placement in `docs/resources/`, correct releasing `feat:` type, and good mock-test coverage. However, it is **not mergeable as-is** — CI is red, and there are three genuine bugs plus several smaller issues.

## Bugs

### 1. CI is red: `TestAccMockDomainContactsImport` fails on the OpenTofu leg (blocking)

The `Mock acceptance (opentofu 1.12.3)` job fails: *"registry.opentofu.org does not have a provider named registry.opentofu.org/hashicorp/namecheap"*. The other two matrix legs (terraform 1.5.7 / 1.15.5) pass.

Root cause: this is the only test whose **first step combines `Config` with `ImportState: true`**. For that shape, terraform-plugin-testing spins up a fresh import working directory, and provider resolution there does not pick up the `TF_ACC_PROVIDER_HOST=registry.opentofu.org` / `TF_ACC_PROVIDER_NAMESPACE=namecheap` overrides the workflow sets for OpenTofu (`ci.yml` mock acceptance step) — so `init` falls back to `hashicorp/namecheap` and hits the registry. The lifecycle test's import step passes because it has no `Config` and reuses the previous step's working directory.

**Fix options:** embed an explicit `terraform { required_providers { namecheap = { source = ... } } }` block in the test config (derived from those env vars), or restructure the import test so the import step follows an applied step, as the lifecycle test does.

### 2. "Domain removed → drop from state" branch in Read is dead code; refresh errors forever instead

In `resourceContactsRead` (`namecheap/namecheap_domain_contacts.go`):

```go
resp, err := client.Domains.GetContactsWithContext(ctx, domain)
if err != nil {
    return diagFromClientError(err)
}
if resp == nil || resp.DomainContactsResult == nil {
    data.SetId("")   // "plans a recreate rather than erroring on every refresh"
    return nil
}
```

In SDK v2.7.0, API-level errors (including `2019166` "Domain not found") are converted into a returned `*namecheap.APIError` by `DoXMLWithContext` — so a domain that has been removed from the account takes the `err != nil` path, and the nil-result branch never fires. The actual behavior is the opposite of the comment and the PR description: every refresh/plan errors with "Domain not found" until the user manually runs `terraform state rm`.

**Fix:** `errors.As(err, &apiErr)` and on code `2019166` (and arguably `2016166` "not associated with account") do `data.SetId(""); return nil` — only fall through to `diagFromClientError` for other errors.

### 3. `CustomizeDiff` breaks when registrant values are unknown at plan time

`customizeContactsDiff` guards only against a fully-empty registrant *block*. If the block is present but a field is interpolated from another resource's not-yet-known attribute (e.g. `email_address = namecheap_x.y.some_computed`), `diff.Get("registrant")` returns the unknown field as `""`, and `SetNew(block, registrant)` plans literal empty strings into `tech`/`admin`/`aux_billing`. At apply time the diff re-runs with the real value and produces different final values → Terraform fails with *"Provider produced inconsistent final plan"*.

**Fix:** check `diff.NewValueKnown("registrant")`; when unknown, use `SetNewComputed(block)` instead of `SetNew`.

Related edge case in the same loop: `cfg.LengthInt()` **panics** on an unknown cty value — reachable with `dynamic "tech" { for_each = <unknown> }`. Guard with `cfg.IsKnown()` before calling `LengthInt()`.

### 4. Live acceptance test permanently clobbers the shared test domain's WHOIS contacts

`TestAccNamecheapDomainContacts` sets fake "Jane Doe" contacts on `NAMECHEAP_TEST_DOMAIN` and — since destroy is state-only by design — **never restores the previous values**. Every run permanently overwrites the shared sandbox domain's contact data. The test should capture the current contacts via `getContacts` up front and restore them in a `CheckDestroy`/`t.Cleanup`.

## Minor issues

- **Import ID not normalized**: the importer lowercases `domain` but leaves `data.Id()` as typed. Importing `EXAMPLE.COM` leaves an uppercase ID that silently flips to lowercase on the first update (`setDomainContacts` calls `SetId(strings.ToLower(...))`). Add `data.SetId(strings.ToLower(data.Id()))` in the importer.
- **Potential perma-diff on `country` case**: the validator accepts lowercase (`pt`), but if Namecheap canonicalizes to uppercase, Read stores `PT` and every plan shows a diff. Either require uppercase in the validator or add a case-insensitive `DiffSuppressFunc`. (The mock stores values verbatim, so mock tests can't catch this — worth a live check.)
- **`setContacts` response `IsSuccess` is ignored**: if the API can return `Status="OK"` with `IsSuccess="false"`, the provider would report success and write state. Cheap to check.
- **Drift test mutates mock state without the lock**: `PreConfig` writes to the map returned by `m.state()` without holding `m.mu`. Sequential in practice, but it undermines the mock's locking discipline (`-race` could flag it if a request ever overlaps).
- **codecov/project is failing** — worth a look, though `codecov/patch` passes, so the new code itself is covered.

## What's good (keep as-is)

- The `contactSchemaFields` single-source-of-truth + `TestContactSchemaCompleteness` reflection test is a great guard against silent data loss on future SDK field additions.
- Defaulting to registrant in `CustomizeDiff` (visible in the plan) is the right call, and the "stable re-plan" mock test proves convergence.
- Delete-as-abandon with a warning diagnostic is the correct semantics for an API with no delete, and it is honestly documented.
- Docs placement (`docs/resources/`, registry-published) and the `feat:` releasing commit type both follow the repo's conventions.

## Verdict

**Request changes.** Items 1–2 are blocking (red CI; Read behavior contradicts documented semantics), item 3 will bite real configs that interpolate contact values, and item 4 should be fixed before anyone runs the live test against a domain they care about.

dm1tr1yvovk (changes requested):
check above

kurok:
@dm1tr1yvovk thanks for the thorough review — addressed and rebased onto master (which now has #284, so the conflict is resolved and the branch is mergeable again). Commits `5361c44`, `e2740e6`, `e993dcc`.

**Bugs**
1. **OpenTofu CI leg** ✅ — `TestAccMockDomainContactsImport` restructured to apply-then-import (no `Config` on the import step), matching the other import tests; passes on all three matrix legs.
2. **Read "domain removed → drop from state" was dead code** ✅ — you're right, the SDK returns `2019166`/`2016166` as an `*APIError`, so the nil-result branch never fired. Read now matches those codes via `errors.As`, clears the ID and returns nil (recreate), and only other errors fall through to `diagFromClientError`.
3. **CustomizeDiff with unknown values** ✅ — now checks `NewValueKnown("registrant")` and uses `SetNewComputed` for the optional blocks when the registrant (or a block) is unknown, and guards `LengthInt` with `IsKnown()` for the unknown-`for_each` dynamic-block case. No more empty-string plans / "inconsistent final plan" / cty panic.
4. **Live test clobbering the shared domain** ✅ — it now captures the domain's current contacts up front and restores them in a `t.Cleanup` that runs after the state-only destroy.

**Minor**
- Import ID normalization ✅ (importer now lowercases `data.Id()` too).
- `setContacts` `IsSuccess=false` now surfaces as an error ✅.
- codecov/project ✅ — added httptest-based unit tests for the read/create/delete + domain-gone paths (coverage back to ~88%, within threshold).
- **Deferred (noted, not blocking):** the `country` case perma-diff depends on whether the API canonicalizes case — the mock can't confirm it, so I'd rather verify against the live sandbox before adding an uppercase `StateFunc`/`DiffSuppressFunc` than guess. The drift-test `PreConfig` lock nit is cosmetic (resource.Test steps are sequential, no real overlap); happy to add a locked mock setter if you'd prefer. Let me know if you want either done now.

CI re-running on the rebased head; `make build/format/check/lint/test/testacc-mock` all green locally.

dm1tr1yvovk (approved):
Approving — verified all four fixes in the code, not just the reply:

1. Import test restructured to apply-then-import; all three mock-acceptance matrix legs (incl. OpenTofu) are green.
2. `isDomainGoneError` correctly matches `*namecheap.APIError` 2019166/2016166 via `errors.As`; the domain-gone path is now live and other errors still surface.
3. `CustomizeDiff` now handles unknown registrant values with `NewValueKnown` + `SetNewComputed`, and `LengthInt` is guarded with `IsKnown()` — both the inconsistent-final-plan and the cty-panic paths are closed.
4. Live test captures and restores the shared domain's contacts in `t.Cleanup`.

Import-ID normalization, the `IsSuccess=false` guard, and the httptest-based coverage tests all look good too.

Both deferrals are fine as-is:
- `country` case: agreed — verify against the live sandbox whether the API canonicalizes case before adding a `DiffSuppressFunc`/`StateFunc`. Please file a follow-up issue so it isn't lost.
- Drift-test `PreConfig` lock: cosmetic, no need to change it now.

Thanks for the thorough turnaround.

PR #286, opened 2026-07-09
feat: add namecheap_domain, namecheap_domains, namecheap_domain_records data sources

### LINKED ISSUE #254: Data sources: namecheap_domain, namecheap_domains (portfolio), namecheap_domain_records
## Problem

The provider has **zero data sources**. Every read-only composition pattern is impossible today:

- reference an existing (non-TF-managed) domain's nameservers or expiry in other resources;
- iterate the account's portfolio (`for_each` over all domains matching a filter) to apply uniform records/policies;
- read the current record set of a domain to merge/augment in other modules;
- gate logic on domain properties (is it expiring? is it locked? which DNS type is active?).

Best part: the SDK **already supports all of it** (`domains.getList`, `domains.getInfo`, `domains.dns.getHosts`, `domains.dns.getList`) — like the quick-win resources, this is pure provider work with no SDK dependency.

## Proposal

Three framework-native data sources:

### 1. `data "namecheap_domain"`

```hcl
data "namecheap_domain" "main" { domain = "example.com" }
# exports: created/expires (RFC3339), is_expired, is_locked, auto_renew, whois_guard,
#          dns_provider_type, nameservers, is_premium, is_our_dns
```

### 2. `data "namecheap_domains"` (portfolio)

```hcl
data "namecheap_domains" "all" {
  search_term = ""          # optional filter, maps to getList params
  list_type   = "ALL"       # ALL | EXPIRING | EXPIRED
}
# exports: domains = list(object({ name, expires, is_locked, auto_renew, whois_guard, ... }))
```

- Auto-paginates via getList (all pages; SDK iterator go-namecheap-sdk#120 when available, manual paging until then — not blocked).
- The killer pattern documented front and center: `for_each = { for d in data.namecheap_domains.all.domains : d.name => d }` → uniform SPF/DMARC records across the whole portfolio in ~10 lines.

### 3. `data "namecheap_domain_records"`

```hcl
data "namecheap_domain_records" "current" { domain = "example.com" }
# exports: records = list(object({ hostname, type, address, ttl, mx_pref })), nameservers, email_type
```

- Read-only view of the live record set (same field shapes as the resources, so outputs compose into resource inputs without transformation).

## Acceptance criteria

- [ ] All three data sources read correctly against mock fixtures (happy path + not-found → clear diagnostic naming the domain, empty portfolio, multi-page portfolio with call-count assertion proving full pagination).
- [ ] Field parity: data-source record/domain object shapes match the corresponding resource schemas attribute-for-attribute (compile-time shared structs + a parity test) so compositions don't need field remapping.
- [ ] Date fields exported as RFC3339 strings + `is_expired`/`expires_in_days` conveniences (timezone-safe; tested at boundaries).
- [ ] The portfolio-wide `for_each` pattern works end-to-end as an acceptance test (data source → 3 mock domains → 3 planned record resources).
- [ ] `list_type` / `search_term` filters map to documented getList params (cross-checked against docs/namecheap-api-v2.md via the SDK).
- [ ] Registry docs for all three with the portfolio-pattern guide; examples/ entries.
- [ ] `make build format check lint test testacc` green; conventional commits; DCO.

## Out of scope

- Pricing/balance data sources (separate issue — needs go-namecheap-sdk#117).
- TLD list data source (defer until a concrete use case; note in epic).
- Caching across plan/apply (Terraform's own data-source semantics apply).

Depends on: framework migration (B), testing infrastructure. **No SDK dependency — ships in the first wave** with the quick-win resources. Part of the provider 3.0 roadmap (see tracking issue).

kurok:
## What & why

The provider had **zero data sources**, blocking every read-only composition pattern (reference an existing domain's nameservers/expiry, iterate the portfolio with `for_each`, merge an existing record set, gate on domain properties). This adds the first three data sources using the existing SDKv2 (`terraform-plugin-sdk/v2`) and `go-namecheap-sdk/v2 v2.7.0` — no SDK bump, pure provider work. `DataSourcesMap` is added to `Provider()`.

## Data sources & exact SDK methods

| Data source | SDK method(s) | Exports |
|---|---|---|
| `namecheap_domain` | `Domains.GetInfoWithContext` (`namecheap.domains.getInfo`) **+** `Domains.GetListWithContext` (`namecheap.domains.getList`) | `dns_provider_type`, `is_our_dns`, `is_premium`, `is_premium_dns`, `nameservers` (from getInfo); `created`, `expires`, `expires_in_days`, `is_expired`, `is_locked`, `auto_renew`, `whois_guard` (from getList) |
| `namecheap_domains` | `Domains.GetListWithContext` (`namecheap.domains.getList`), auto-paginated across all pages | `domains = list(object({ id, name, user, created, expires, expires_in_days, is_expired, is_locked, auto_renew, whois_guard, is_premium, is_our_dns }))`; filters `search_term`, `list_type` (ALL/EXPIRING/EXPIRED) |
| `namecheap_domain_records` | `DomainsDNS.GetHostsWithContext` + `DomainsDNS.GetListWithContext` | `records = list(object({ hostname, type, address, mx_pref, ttl }))`, `nameservers`, `email_type` |

## `namecheap_domain` exports the full spec (two-call design)

`namecheap.domains.getInfo` returns only DNS-oriented data; it does **not** return the domain lifecycle fields. Rather than drop those fields, `namecheap_domain` now issues a second `namecheap.domains.getList` call, `SearchTerm`-narrowed and exact-matched client-side (SearchTerm is a substring filter, not an exact lookup), and copies the lifecycle fields via the shared `flattenPortfolioDomain` helper. `getInfo` remains the authoritative existence check (its not-found diagnostic names the domain), so a portfolio miss leaves the lifecycle fields at their zero values rather than failing the read. The two-call cost is documented on the registry page; callers needing lifecycle data for many domains at once should prefer the one-listing `namecheap_domains` portfolio source.

## Field parity

The `records` element schema is produced by a shared builder (`domainRecordElemSchema`) and matches the `namecheap_domain_records` resource `record` block attribute-for-attribute, so outputs feed into resource inputs (incl. a `dynamic "record"` block) with no remapping. Enforced by `TestDomainRecordFieldParity`.

## Dates

`created`/`expires` are exported as RFC3339 UTC strings; `expires_in_days` is a timezone-safe whole-calendar-day convenience (negative when expired). `is_expired` comes straight from the API. Boundary behaviour is unit-tested (`TestFormatDateTime`, `TestDaysUntil`, `TestDaysUntilTimezoneSafe`).

## Testing performed

- `gofmt -l namecheap/` clean, `go build ./...`, `go vet ./...` (incl. `-tags testacc`), `golangci-lint run` → 0 issues, `make build` OK.
- **Unit tests (`go test ./namecheap/...`, no build tag → observed by the coverage upload):** data-source registration, record field parity, date/`expires_in_days` helpers incl. timezone boundaries, **plus new `data_source_read_test.go`** that drives all three `ReadContext` functions against an in-process `httptest` server through the real SDK client (`client.BaseURL`), covering success, not-found (diagnostic names the domain), empty portfolio, multi-page pagination, and API-error paths. Project coverage **83.6% → 91.5%** (above the pre-PR baseline), fixing the `codecov/project` gate that the earlier revision tripped (the mock suite below is `testacc`-gated and therefore invisible to codecov).
- Mock-backed acceptance tests (`make testacc-mock`, extends the existing stateful mock with `getInfo`/`getList`): happy path for all three (the `namecheap_domain` case now asserts the getList-sourced lifecycle fields end-to-end); **not-found → diagnostic naming the domain**; **empty, single-page, and multi-page portfolio with a getList call-count assertion proving full pagination**; and an end-to-end **portfolio → 3 mock domains → 3 record resources** composition (asserted against the mock backend + cleared on destroy).
  - Note: the acceptance test drives that composition with `count` rather than `for_each`, because the SDKv2 acceptance harness (`terraform-plugin-testing`'s legacy state shim) cannot represent string-keyed `for_each` instances. The identical data-source-driven composition is exercised; the idiomatic `for_each` form is documented as the recommended user pattern in `docs/data-sources/domains.md`.

## Docs

Registry docs for all three under `docs/data-sources/`, including the portfolio-wide `for_each` guide and a merge/augment `dynamic` block example. `docs/data-sources/domain.md` documents the two-call cost and the full attribute set.

## Not done (out of scope per issue)

Pricing/balance and TLD-list data sources are explicitly out of scope. There is no `examples/` tree in this repo, so runnable examples live inline in the `docs/data-sources/` pages (matching the existing docs-only layout).

Closes #254

### DIFF
--- a/namecheap/data_source_helpers.go
+++ b/namecheap/data_source_helpers.go
@@ -1,0 +1,208 @@
+package namecheap_provider
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/namecheap/go-namecheap-sdk/v2/namecheap"
+)
+
+// dataSourceDomainReadError converts an SDK error from a domain-scoped read into
+// diagnostics that name the domain. It reuses diagFromClientError so that known
+// Namecheap error codes (e.g. 2019166 "Domain not found") keep their remediation
+// text, while the summary is prefixed with the domain so a failed lookup clearly
+// identifies which domain could not be read.
+func dataSourceDomainReadError(domain string, err error) diag.Diagnostics {
+	diags := diagFromClientError(err)
+	for i := range diags {
+		diags[i].Summary = fmt.Sprintf("%s (domain %q)", diags[i].Summary, domain)
+	}
+	return diags
+}
+
+// formatDateTime renders a Namecheap DateTime as an RFC3339 string. Namecheap
+// returns lifecycle dates as calendar dates (the SDK parses "MM/DD/YYYY" with
+// time.Parse, yielding midnight UTC), so the RFC3339 output is timezone-stable.
+// A nil pointer renders as an empty string.
+func formatDateTime(dt *namecheap.DateTime) string {
+	if dt == nil {
+		return ""
+	}
+	return dt.Time.UTC().Format(time.RFC3339)
+}
+
+// daysUntil returns the whole number of calendar days from now until target,
+// negative when target is in the past. Both instants are reduced to their UTC
+// calendar date before subtracting, so the result is independent of the wall
+// clock time of day and of the local timezone. now is passed explicitly so the
+// boundary behaviour is deterministically testable.
+func daysUntil(target, now time.Time) int {
+	targetDay := time.Date(target.UTC().Year(), target.UTC().Month(), target.UTC().Day(), 0, 0, 0, 0, time.UTC)
+	nowDay := time.Date(now.UTC().Year(), now.UTC().Month(), now.UTC().Day(), 0, 0, 0, 0, time.UTC)
+	return int(targetDay.Sub(nowDay).Hours() / 24)
+}
+
+// derefString returns the dereferenced value of a *string, or "" when nil.
+func derefString(p *string) string {
+	if p == nil {
+		return ""
+	}
+	return *p
+}
+
+// derefBool returns the dereferenced value of a *bool, or false when nil.
+func derefBool(p *bool) bool {
+	if p == nil {
+		return false
+	}
+	return *p
+}
+
+// derefInt returns the dereferenced value of a *int, or 0 when nil.
+func derefInt(p *int) int {
+	if p == nil {
+		return 0
+	}
+	return *p
+}
+
+// domainRecordElemSchema returns the schema for a single DNS host record as
+// exported by the namecheap_domain_records data source. The attribute names and
+// types intentionally mirror the record block of the namecheap_domain_records
+// resource (see resourceNamecheapDomainRecords), so a data-source record object
+// feeds straight into a resource record block without field remapping. This
+// parity is asserted by TestDomainRecordFieldParity.
+func domainRecordElemSchema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"hostname": {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "Sub-domain/hostname of the record.",
+		},
+		"type": {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "Record type (e.g. A, AAAA, CNAME, MX, TXT).",
+		},
+		"address": {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "Record value (URL or IP address, depending on the record type).",
+		},
+		"mx_pref": {
+			Type:        schema.TypeInt,
+			Computed:    true,
+			Description: "MX preference for the host. Applicable to MX records only.",
+		},
+		"ttl": {
+			Type:        schema.TypeInt,
+			Computed:    true,
+			Description: "Time to live for the record, in seconds.",
+		},
+	}
+}
+
+// flattenHostRecord converts an SDK detailed host record into the map shape
+// described by domainRecordElemSchema.
+func flattenHostRecord(h *namecheap.DomainsDNSHostRecordDetailed) map[string]interface{} {
+	return map[string]interface{}{
+		"hostname": derefString(h.Name),
+		"type":     derefString(h.Type),
+		"address":  derefString(h.Address),
+		"mx_pref":  derefInt(h.MXPref),
+		"ttl":      derefInt(h.TTL),
+	}
+}
+
+// domainPortfolioElemSchema returns the schema for a single domain as exported
+// by the namecheap_domains portfolio data source.
+func domainPortfolioElemSchema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"id": {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "Namecheap internal domain identifier.",
+		},
+		"name": {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "The domain name (e.g. example.com).",
+		},
+		"user": {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "The account user the domain belongs to.",
+		},
+		"created": {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "Registration date as an RFC3339 timestamp (UTC).",
+		},
+		"expires": {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "Expiration date as an RFC3339 timestamp (UTC).",
+		},
+		"expires_in_days": {
+			Type:        schema.TypeInt,
+			Computed:    true,
+			Description: "Whole calendar days until the domain expires (negative if already expired).",
+		},
+		"is_expired": {
+			Type:        schema.TypeBool,
+			Computed:    true,
+			Description: "Whether the domain has expired.",
+		},
+		"is_locked": {
+			Type:        schema.TypeBool,
+			Computed:    true,
+			Description: "Whether the registrar lock is enabled.",
+		},
+		"auto_renew": {
+			Type:        schema.TypeBool,
+			Computed:    true,
+			Description: "Whether auto-renew is enabled.",
+		},
+		"whois_guard": {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "WhoisGuard status (e.g. ENABLED, DISABLED, NOTPRESENT).",
+		},
+		"is_premium": {
+			Type:        schema.TypeBool,
+			Computed:    true,
+			Description: "Whether the domain is a premium domain.",
+		},
+		"is_our_dns": {
+			Type:        schema.TypeBool,
+			Computed:    true,
+			Description: "Whether the domain is using Namecheap's DNS.",
+		},
+	}
+}
+
+// flattenPortfolioDomain converts an SDK Domain (from getList) into the map
+// shape described by domainPortfolioElemSchema. now is used to compute
+// expires_in_days.
+func flattenPortfolioDomain(d *namecheap.Domain, now time.Time) map[string]interface{} {
+	m := map[string]interface{}{
+		"id":              derefString(d.ID),
+		"name":            derefString(d.Name),
+		"user":            derefString(d.User),
+		"created":         formatDateTime(d.Created),
+		"expires":         formatDateTime(d.Expires),
+		"expires_in_days": 0,
+		"is_expired":      derefBool(d.IsExpired),
+		"is_locked":       derefBool(d.IsLocked),
+		"auto_renew":      derefBool(d.AutoRenew),
+		"whois_guard":     derefString(d.WhoisGuard),
+		"is_premium":      derefBool(d.IsPremium),
+		"is_our_dns":      derefBool(d.IsOurDNS),
+	}
+	if d.Expires != nil {
+		m["expires_in_days"] = daysUntil(d.Expires.Time, now)
+	}
+	return m
+}
--- a/namecheap/data_source_helpers.go
+++ b/namecheap/data_source_helpers.go
@@ -1,13 +1,61 @@
 package namecheap_provider
 
 import (
+	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/namecheap/go-namecheap-sdk/v2/namecheap"
 )
+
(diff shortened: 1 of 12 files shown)


kurok:
### Re: the red `Stop self-hosted EC2 runner` check

This is a **CI runner-teardown failure, not a code or test failure** — safe to disregard for the purpose of judging this PR.

**What actually happened** (run [`29005677212`](https://github.com/namecheap/terraform-provider-namecheap/actions/runs/29005677212)):

- ✅ The `Acceptance test` job (live sandbox suite) **passed**.
- ❌ The subsequent `Stop self-hosted EC2 runner` job hung for ~15 min and was **cancelled** (`conclusion: cancelled`, surfaced as `fail` in the checks list).
- I re-ran the failed job once (`gh run rerun --failed`); it cancelled the same way. So it's **not a transient, re-runnable flake** — the stop-into-pool step itself is wedging, independent of this branch.

**Why it's unrelated to this change:** the diff is Go-only (data-source read logic + unit/mock tests + docs). It cannot influence an AWS EC2 instance-teardown job, which runs *after* the tests and only tears down the warm-pool runner. Any orphaned instance is reclaimed by the runner leak reaper / TTL (see the warm-pool lifecycle in #281), so there's no lingering-cost concern.

**Everything code-relevant is green**, including the check this PR revision was specifically fixing:

| Check | Result |
|---|---|
| `codecov/project` | ✅ pass (was failing pre-fix; project coverage 83.6% → 91.5%) |
| `codecov/patch` | ✅ pass |
| `Acceptance test` (live sandbox) | ✅ pass |
| `Mock acceptance` (terraform 1.5.7 / 1.15.5, opentofu 1.12.3) | ✅ pass |
| `Check`, `CodeQL`, `DCO`, `Trivy`, `govulncheck`, PR title, Snyk | ✅ pass |
| `Stop self-hosted EC2 runner` | ❌ cancelled (infra teardown — see above) |

Recommend not `--admin`-overriding it; if a fully clean board is required before merge, the fix belongs on the runner-workflow / AWS side (the stop step), not here.

dm1tr1yvovk:
### Code review

Found 1 issue:

1. The `namecheap_domain_records` data source returns Namecheap's default parking records unfiltered ([CLAUDE.md](https://github.com/namecheap/terraform-provider-namecheap/blob/9c736ccfabb1b4df20634a485c9b05028bfb6943/CLAUDE.md#L106-L108) says "Default parking records (CNAME www→parkingpage.namecheap.com, URL @→http://www.domain) are returned by the API and must be filtered."). The read loop appends every host from `GetHostsWithContext` verbatim, while the resource read path filters them via `filterDefaultParkingRecords` ([namecheap_domain_record_functions.go#L273](https://github.com/namecheap/terraform-provider-namecheap/blob/9c736ccfabb1b4df20634a485c9b05028bfb6943/namecheap/namecheap_domain_record_functions.go#L272-L274)). Since `docs/data-sources/domain_records.md` documents feeding `records` into a `namecheap_domain_records` resource via a `dynamic` block, a parked domain would write the parking records back into the zone and produce perpetual plan drift. Fix: apply `filterDefaultParkingRecords` to `hostsResp.DomainDNSGetHostsResult.Hosts` before flattening.

https://github.com/namecheap/terraform-provider-namecheap/blob/9c736ccfabb1b4df20634a485c9b05028bfb6943/namecheap/data_source_namecheap_domain_records.go#L84-L89

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

namecheap/data_source_namecheap_domain_records.go
```diff
@@ -0,0 +1,96 @@
...
+
+	nameservers := []string{}
+	if !*nsResp.DomainDNSGetListResult.IsUsingOurDNS && nsResp.DomainDNSGetListResult.Nameservers != nil {
+		nameservers = *nsResp.DomainDNSGetListResult.Nameservers
+	}
+	if err := data.Set("nameservers", nameservers); err != nil {
+		return diag.FromErr(err)
+	}
+
+	hostsResp, err := client.DomainsDNS.GetHostsWithContext(ctx, domain)
```
dm1tr1yvovk, line 91: `GetHostsWithContext` is called unconditionally, even when `IsUsingOurDNS` is false. The resource explicitly branches away from `GetHosts` for custom-NS domains — see the comment in `resourceRecordRead` (namecheap_domain_record.go:224): reading records fails when Namecheap doesn't control the zone. The mock test (`TestDataSourceDomainRecordsRead_CustomNS`) assumes the API succeeds with an empty set, but the real API may error. Suggest guarding this call on `IsUsingOurDNS` and returning nameservers + empty `records` otherwise, mirroring the resource.

namecheap/data_source_helpers.go
```diff
@@ -0,0 +1,256 @@
...
+			Type:        schema.TypeInt,
+			Computed:    true,
+			Description: "Time to live for the record, in seconds.",
+		},
+	}
+}
+
+// flattenHostRecord converts an SDK detailed host record into the map shape
+// described by domainRecordElemSchema.
+func flattenHostRecord(h *namecheap.DomainsDNSHostRecordDetailed) map[string]interface{} {
```
dm1tr1yvovk, line 206: `flattenHostRecord` returns the raw API address without the normalization the resource read path applies via `getFixedAddressOfRecord` (trailing dot for CNAME/ALIAS/NS/MX, quoting for CAA — see `readRecordsMerge`/`readRecordsOverwrite`). Feeding these records into a `namecheap_domain_records` resource per the docs example will plan-drift on those record types because the resource stores the normalized form. Suggest applying `getFixedAddressOfRecord` here (or reusing `convertDomainRecordDetailedToTypeSetRecord`) so both paths emit identical values.

namecheap/data_source_helpers.go
```diff
@@ -0,0 +1,256 @@
...
+// the exact domain is matched client-side (case-insensitively) among the
+// returned rows. A transport/API error is surfaced (named with the domain); a
+// clean response that simply does not contain the domain is treated as "no
+// lifecycle data available" and leaves the fields at their zero values, because
+// the caller has already confirmed the domain exists via getInfo.
+func setDomainLifecycleFromList(ctx context.Context, client *namecheap.Client, data *schema.ResourceData, domain string) diag.Diagnostics {
+	resp, err := client.Domains.GetListWithContext(ctx, &namecheap.DomainsGetListArgs{
+		SearchTerm: namecheap.String(domain),
+		Page:       namecheap.Int(1),
+		PageSize:   namecheap.Int(domainsPageSize),
```
dm1tr1yvovk: `setDomainLifecycleFromList` fetches only page 1 (`PageSize: 100`). `SearchTerm` is a server-side substring filter, so on an account with 100+ domains containing the queried name as a substring, the exact match can land on page 2+ and every lifecycle field (`created`, `expires`, `expires_in_days`, `is_locked`, `auto_renew`, `whois_guard`) silently stays at its zero value — even though `getInfo` just confirmed the domain exists. `dataSourceNamecheapDomainsRead` in this same PR already implements the paging loop; suggest reusing it here, or at minimum emitting a warning diagnostic on a miss instead of degrading silently.

dm1tr1yvovk (changes requested):
Requesting changes based on one confirmed bug (parking-record filtering, see [earlier comment](https://github.com/namecheap/terraform-provider-namecheap/pull/286#issuecomment-4925190966)) plus three read-path parity gaps flagged inline below.

Common theme: the data sources re-implement the read path instead of reusing the workarounds the `namecheap_domain_records` resource already encodes (`filterDefaultParkingRecords`, `getFixedAddressOfRecord`, the custom-NS guard). Since the docs explicitly promote composing data-source output into the resource, the two paths need to normalize identically or the documented pattern produces drift.

None of these require design changes — each is a small fix reusing existing helpers.

kurok:
@dm1tr1yvovk thanks — all four addressed by reusing the resource's read-path helpers, and rebased onto master (now with #284, so the branch is mergeable again). Commit `a2a24fe`.

- **Parking records** ✅ — `namecheap_domain_records` now runs `filterDefaultParkingRecords` on the getHosts result before flattening, so a parked domain no longer round-trips parking records into the resource.
- **Custom-NS guard** ✅ — GetHosts is now called only when `IsUsingOurDNS`; a custom-nameserver domain returns its nameservers + an empty record set (and empty `email_type`), mirroring `resourceRecordRead` and avoiding the GetHosts error the real API can raise.
- **Address normalization** ✅ — `flattenHostRecord` now normalizes via `getFixedAddressOfRecord` (trailing dot for CNAME/ALIAS/NS/MX, quotes for CAA), falling back to the raw value on a malformed field, so exported records match what the resource stores.
- **Lifecycle pagination** ✅ — `setDomainLifecycleFromList` now walks the entire getList result via a shared `fetchAllDomains` helper (also used by the portfolio read, so the paging logic is single-sourced) instead of only page 1, and emits a warning diagnostic if the domain is genuinely absent from the listing rather than degrading silently.

`make build/format/check/lint/test/testacc-mock` all green; coverage 91.1%. CI re-running on the rebased head.

namecheap/data_source_namecheap_domain_records.go
kurok: Fixed in a2a24fe — GetHosts is now guarded on IsUsingOurDNS; a custom-nameserver domain returns its nameservers + an empty record set (no GetHosts call), mirroring resourceRecordRead.

namecheap/data_source_helpers.go
kurok: Fixed in a2a24fe — flattenHostRecord now normalizes the address via getFixedAddressOfRecord (trailing dot for CNAME/ALIAS/NS/MX, quotes for CAA), falling back to the raw value on a malformed field, so it matches what the resource stores.

namecheap/data_source_helpers.go
kurok: Fixed in a2a24fe — setDomainLifecycleFromList now pages the entire getList result via a shared fetchAllDomains helper (also used by the portfolio read), and emits a warning if the domain is genuinely absent rather than degrading silently.

dm1tr1yvovk (approved):
All four findings from the previous review are addressed in a2a24fe and verified against the diff:

- Parking records filtered via `filterDefaultParkingRecords` before flattening, matching the resource read path.
- `GetHosts` now guarded on `IsUsingOurDNS`; custom-NS domains return nameservers + empty record set.
- `flattenHostRecord` normalizes addresses via `getFixedAddressOfRecord`, so data-source output composes into the resource without drift.
- Lifecycle lookup pages the full `getList` result via the shared `fetchAllDomains` helper and warns on a genuine portfolio miss instead of silently zeroing fields.

CI green on the rebased head (unit, mock acceptance, live acceptance, codecov project/patch). Approving.

Issue #260, opened 2026-07-03
readRecordsOverwrite does not filter default parking records (OVERWRITE-mode drift)


dm1tr1yvovk:
### Summary

`readRecordsOverwrite` does not filter out Namecheap's default parking records, so they leak into Terraform state as spurious drift in OVERWRITE mode.

### Details

`createRecordsMerge` correctly filters default parking records via `filterDefaultParkingRecords` before processing the remote host set:

https://github.com/namecheap/terraform-provider-namecheap/blob/fff6bf88c07664e4d703143c80cac736480f3f80/namecheap/namecheap_domain_record_functions.go#L272-L285

However, `readRecordsOverwrite` appends **every** remote record to the result without filtering:

https://github.com/namecheap/terraform-provider-namecheap/blob/fff6bf88c07664e4d703143c80cac736480f3f80/namecheap/namecheap_domain_record_functions.go#L406-L429

The `filterDefaultParkingRecords` helper already exists for exactly this purpose:

https://github.com/namecheap/terraform-provider-namecheap/blob/fff6bf88c07664e4d703143c80cac736480f3f80/namecheap/namecheap_domain_record_functions.go#L681-L694

### Impact

On a parked or freshly-imported domain, the default parking records (`CNAME www -> parkingpage.namecheap.com.`, `URL @ -> http://www.<domain>`) are read into state and show up as unexpected drift during `terraform plan` in OVERWRITE mode. This contradicts the documented invariant that default parking records "are returned by the API and must be filtered" (see the SDK gotchas in `CLAUDE.md`).

Note: `readRecordsMerge` only returns records that match the user's config, so it does not exhibit the leak; the fix is specifically needed in `readRecordsOverwrite`.

### Suggested fix

Apply `filterDefaultParkingRecords(...)` to `remoteRecordsResponse.DomainDNSGetHostsResult.Hosts` in `readRecordsOverwrite` before iterating, mirroring `createRecordsMerge`.

### Context

Surfaced during review of #258 (SDK v2.7.0 / WithContext migration). Pre-existing issue, not introduced by that PR; filed separately so the migration PR stays purely mechanical.

Issue #250, opened 2026-07-02
OVERWRITE safety: plan-time warnings for unmanaged-record deletion + allow_unmanaged_deletion consent knob (#65)


kurok:
## Problem

Open issue #65: `namecheap_domain_records` in OVERWRITE mode deletes every record it doesn't know about — including records created manually or by other tooling — **without any warning in the plan**. The user sees "1 to change" and loses their manually-added records on apply. This is the provider's live data-loss foot-gun, and it stays relevant even after the per-record resource ships (full-ownership mode remains the right model for zone-file-style management).

## Proposal

Make the destruction visible and consentable — three graduated mechanisms:

### 1. Plan-time removal visibility

- During plan/refresh, diff the *live* record set against state + config; records that exist upstream but are absent from config surface as **warning diagnostics enumerating exactly what OVERWRITE will delete**:

  > Warning: OVERWRITE will delete 3 records not in configuration: `A blog → 1.2.3.4`, `TXT @ → "v=spf1 …"`, `MX @ → mail.example.com (10)`

- Implemented in the existing refresh path (live read already happens); costs no extra API calls.

### 2. Explicit consent knob

```hcl
resource "namecheap_domain_records" "main" {
  mode  = "OVERWRITE"
  allow_unmanaged_deletion = false   # NEW, default false
  ...
}
```

- Default `false`: plan **errors** (not warns) when OVERWRITE would delete unmanaged records — the safe default; setting `true` restores current semantics with the warning from (1).
- This is a behavior change for silent-delete users → ships in the 3.0 major (framework migration phase C) with a prominent migration-guide entry; until then, (1)'s warnings ship immediately in a minor as pure addition.

### 3. Drift-adoption escape hatch

- The warning text includes copy-paste HCL record blocks for each about-to-be-deleted record, so "actually I want to keep those" is a paste, not an archaeology session.

## Acceptance criteria

- [ ] Warning diagnostics list every to-be-deleted unmanaged record with type/host/value (acceptance test: seed mock with extra records → plan → assert diagnostic contents; apply proceeds).
- [ ] `allow_unmanaged_deletion = false` (default, 3.0) turns the case into a plan-time **error** naming the knob; `true` restores apply-with-warning (both acceptance-tested).
- [ ] MERGE mode unaffected (regression tests); no extra API calls added to plan (mock call-count assertion).
- [ ] Emitted HCL suggestions in the warning are valid HCL that, when pasted, produce an empty diff (golden test).
[...]

Issue #245, opened 2026-07-02
Migrate to terraform-plugin-framework via mux transition (SDKv2 → framework)


kurok:
## Problem

The provider is built on `terraform-plugin-sdk/v2` — HashiCorp's legacy SDK, in maintenance mode since 2022. Staying on it costs us concretely:

- **No nested-attribute schemas, plan modifiers, or custom types** — the exact tools the roadmap's new resources (domain lifecycle, per-record DNS) need for good UX (e.g. `RequiresReplace` on the right attributes, defaults without `DiffSuppressFunc` hacks).
- **Weaker validation**: SDKv2 validates at apply-time in many paths; the framework validates at plan time with attribute-level diagnostics.
- **Protocol v5 only** — excluded from newer Terraform features (moved blocks for resource-type moves, config-driven import improvements, ephemeral values interop).
- Every new resource we add to SDKv2 grows the eventual migration; the roadmap adds ~8 resources/data sources, so the cheapest moment to switch is **now**, before the coverage wave.

## Proposal

Adopt `terraform-plugin-framework` via the standard **mux transition** — no big-bang rewrite:

1. **Phase A — mux in place**: serve the existing SDKv2 provider and a new empty framework provider behind `tf5muxserver` (staying protocol v5 while SDKv2 is in the mix), single binary, zero user-visible change. Provider config schema implemented identically on both sides (a mux requirement — schemas must match).
2. **Phase B — new work lands framework-side**: every new roadmap resource/data source (per-record DNS, domain lifecycle, data sources, quick-win resources) is written against the framework from day one. SDKv2 side is frozen except bugfixes.
3. **Phase C — migrate `namecheap_domain_records`**: port the one existing resource (state compatibility verified — same type name, same schema shape, no state upgrade needed if attribute types are preserved; a state-upgrade test proves it), drop SDKv2 + mux, optionally move to protocol v6.

Version/compat policy: phases A+B ship in minor releases (release-please `feat:`); phase C is the 3.0.0 major (framework requires Terraform ≥ 1.0 / protocol v5; document tested Terraform + OpenTofu versions).

## Acceptance criteria

- [ ] Phase A: muxed binary serves the current provider with **zero behavior change** — full existing unit-test suite green, plus a mux smoke test (provider configure + domain_records plan) against both served schemas; provider config schema parity asserted by test.
[...]

PR #288, opened 2026-07-10
feat: warn before OVERWRITE mode deletes unmanaged DNS records

### LINKED ISSUE #65: OVERWRITE Mode does not warn about removing manual records
### Terraform Version
v1.0.4

### Namecheap provider version
v2.0.2

### Affected Resource(s)
Please list the resources as a list, for example:
- namecheap_domain_records

### Terraform Configuration Files
```hcl
resource "namecheap_domain_records" "my-domain-com" {
  domain = "some.domain"
  mode = "OVERWRITE"
  email_type = "NONE"

  record {
    hostname = "hostname"
    type = "CNAME"
    address = "some.address"
  }

}
```

### Debug Output
Terraform will perform the following actions:
```
  # namecheap_domain_records.my-domain-com will be created
  + resource "namecheap_domain_records" "my-domain-com" {
      + domain     = "some.domain"
      + email_type = "NONE"
      + id         = (known after apply)
      + mode       = "OVERWRITE"

      + record {
          + address  = "some.address"
          + hostname = "hostname"
          + mx_pref  = 10
          + ttl      = 1799
          + type     = "CNAME"
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.
```

### Expected Behavior
Should warn about removing existing resources that are created manually. Even thou the manual records are not in the TFState, if the API connection has already been made, and the OVERWRITE flag it true, some warning should be made.

### Actual Behavior
Only says it is adding a resource, not deleting (outside of terraform, but still under the api call).

### Steps to Reproduce

1. `terraform apply`

### Important Factoids
Because of this mistake, I accidentally deleted all production records and had to contact Namecheap support to restore them. It's unclear from the docs that the terraform will manage anything that was not created first by it, as intuitively this is terraform's use case. If mass-deletion was required, it would be better not to provide it using this provider as it can create confusion. If it's necessary, a warning should be made on the docs wherever Overwrite is written.

kurok:
## Summary

`OVERWRITE` mode replaces the entire DNS zone with zero visibility into what gets destroyed: `createRecordsOverwrite`/`deleteRecordsOverwrite` call `SetHosts` with no prior read, and every CRUD entry point discarded non-error diagnostics, so unmanaged-record deletion was always silent (#65).

This is Stage 1 of the plan in #250's [implementation-plan comment](https://github.com/namecheap/terraform-provider-namecheap/issues/250#issuecomment-4933342788) — the pure-additive warning layer. Per that plan, the `allow_unmanaged_deletion` consent knob (Stage 2) is deliberately **not** included here: it needs a true plan-time error, which in SDKv2 would require an extra API call during `plan` (violating this issue's own "no extra API calls to plan" AC) and can't distinguish "unmanaged record surfaced by refresh" from "record deliberately removed from config." That's only implementable cleanly after the `terraform-plugin-framework` migration (#245), where `ModifyPlan` + private state make it free. #250 stays open for that follow-up; this PR closes #65 since it delivers the visibility #65 asked for.

- Refresh-time warning (visible at `terraform plan`) enumerating live records not in the configuration, gated to `OVERWRITE` mode only (never `IMPORT`, which is adopting records, not about to delete them).
- Apply-time pre-flight warning in `createRecordsOverwrite`/`deleteRecordsOverwrite` (one extra `GetHosts` before the destructive `SetHosts`) — this is the only mechanism that can cover the #65 *create* case, since a brand-new resource has no prior state for `plan` to diff against.
- Each warning includes a paste-ready `record { ... }` block per record so an unmanaged record can be adopted instead of lost; adopting it produces an empty subsequent plan (golden-tested).
- Zero additional API calls during plan/refresh; exactly one added `GetHosts` per `OVERWRITE` apply.
- `MERGE` mode, `IMPORT`, and default-parking-record filtering (#260) are unaffected — call counts and outcomes unchanged.
- A pre-flight `GetHosts` failure fails closed (error, no `SetHosts` reached) rather than proceeding to a blind zone wipe.

## Test plan

- [x] `make build format check lint test` — unit suite green, no coverage regression (91.3%)
- [x] `make testacc-mock` — full mock acceptance suite green, including the existing MERGE/OVERWRITE regression matrix (#275-#280) with unchanged call-count expectations
- [x] New unit coverage: unmanaged-set computation (empty/mixed/parking-filtered/CAA-quoting/CNAME-dot-normalization), `formatRecordHCL` golden tests (A/CNAME/MX-with-and-without-pref/TXT-with-quotes/CAA/URL), warning builder (both tenses), CRUD-level warning threading, IMPORT non-warning
- [x] New mock acceptance coverage: create-over-existing (#65 repro), plan adds zero extra `GetHosts` calls, golden HCL adoption round-trip (empty plan after pasting), pre-flight API failure aborts before any destructive write
- [x] Docs updated: `docs/resources/domain_records.md`, `docs/guides/namecheap_domain_records_guide.md`, `docs/index.md`


dm1tr1yvovk (dismissed):
Approving — the warning layer is well built (fail-closed pre-flight, parking-record filtering, IMPORT exclusion, solid test matrix), and CLAUDE.md compliance checks out. One finding worth addressing, either here or in a follow-up:

**Spurious warning on intentional record removal during OVERWRITE updates.** `resourceRecordUpdate` passes the *new* config to `createRecordsOverwrite`, which uses it as the reference set for `unmanagedRecordsOverwrite`:

https://github.com/namecheap/terraform-provider-namecheap/blob/bffc2f9b7e675060ea8bea5e100dc1862e500a3c/namecheap/namecheap_domain_record.go#L381-L383

https://github.com/namecheap/terraform-provider-namecheap/blob/bffc2f9b7e675060ea8bea5e100dc1862e500a3c/namecheap/namecheap_domain_record_functions.go#L340-L348

A record the user just deliberately removed from config is still live at pre-flight time, so every OVERWRITE apply that shrinks the record set fires "OVERWRITE mode will delete N record(s) not present in the configuration" — for a fully consented deletion. Note the destroy path already gets this right by comparing against prior state instead:

https://github.com/namecheap/terraform-provider-namecheap/blob/bffc2f9b7e675060ea8bea5e100dc1862e500a3c/namecheap/namecheap_domain_record_functions.go#L652-L654

Suggestion: on the update path, compute the unmanaged set against prior state (records Terraform already knew about) rather than the incoming config, so only true out-of-band drift is flagged. There's currently no test covering the update-removes-a-record scenario — worth adding one either way.

kurok:
@dm1tr1yvovk Fixed the spurious-warning-on-update finding in bc3a12d.

`createRecordsOverwrite` now takes a `priorRecords` parameter, unioned with the incoming records to form the pre-flight's reference set. The update call site (`namecheap_domain_record.go`) passes `oldRecords`; create passes `nil` (no prior state to consult, matching the existing #65 create-case behavior).

Added `TestCreateRecordsOverwrite_UpdateRemovingRecordDoesNotWarn` and its complement `_UpdateStillWarnsAboutTrueDrift`. Verified the first test actually catches the regression by temporarily disabling the `priorRecords` union locally and confirming it fails with exactly the spurious "will delete api" warning you described, then restored the fix and confirmed both pass.

dm1tr1yvovk (approved):
Re-approving after bc3a12d. The spurious-warning-on-update finding is resolved: the pre-flight reference set now unions the incoming config with prior state (`priorRecords`), so deliberate config removals are treated as consented while true out-of-band drift still warns — covered by the new `TestCreateRecordsOverwrite_UpdateRemovingRecordDoesNotWarn` / `_UpdateStillWarnsAboutTrueDrift` pair. CI is green including acceptance tests, DCO passes.

Issue #249, opened 2026-07-02
Quick-win resources on existing SDK surface: namecheap_email_forwarding (#85) and namecheap_nameserver (#76)


kurok:
## Problem

Two community-requested resources have been waiting on nothing but provider work — the SDK has supported both for years:

- **#85 — email forwarding**: `domains.dns.getEmailForwarding` / `setEmailForwarding` are in go-namecheap-sdk today. Users manage mailbox redirects (`info@example.com → team@gmail.com`) by hand in the dashboard while the rest of the domain is in Terraform.
- **#76 — custom nameservers**: `domains.ns.create/delete/update/getInfo` are in the SDK today. Users running their own DNS (glue records / vanity nameservers like `ns1.example.com`) can't register them via Terraform.

These are the cheapest real value in the roadmap: no SDK dependency, clear community demand, small blast radius.

### 1. `namecheap_email_forwarding` (fixes #85)

```hcl
resource "namecheap_email_forwarding" "info" {
  domain = "example.com"
  forwards = {
    "info"  = "team@example-corp.com"
    "sales" = "sales@example-corp.com"
  }
}
```

- One resource per domain owning the full forwarding table (the API is set-based like setHosts, so full-ownership is the honest model — documented); map attribute keeps diffs readable.
- Interaction constraint surfaced clearly: the API ties email forwarding to Namecheap DNS/MX settings — plan-time validation warns when the domain's records/MX setup conflicts (the classic silent-failure mode).

### 2. `namecheap_nameserver` (fixes #76)

```hcl
resource "namecheap_nameserver" "ns1" {
  domain = "example.com"
  name   = "ns1.example.com"
  ip     = "203.0.113.10"
}
```

- Per-nameserver resource (create/update IP/delete map 1:1 to the API); import supported (`example.com/ns1.example.com`).
- Update = IP change in place (`domains.ns.update`); rename = replace (RequiresReplace on `name`).

Both framework-native (per the migration issue), both with the standard lifecycle acceptance matrix, both shipped with registry docs + examples.

## Acceptance criteria

- [ ] `namecheap_email_forwarding`: create/update (add/change/remove entries)/destroy(clears table)/import/drift matrix green against the mock; a seeded-then-managed test proves the full-ownership semantics are enforced and documented (refresh shows external additions as drift).
- [ ] Email-forwarding ↔ MX conflict detection: acceptance test with conflicting custom MX records produces the plan-time warning (not a silent apply-then-broken-email).
[...]

Issue #85, opened 2024-01-13
Add support for email redirect rules


StefanTheWiz:
Hi

I'd like to be able to set the email redirect rules via Terraform, without having to log into the control panel to do it manually.

* Terraform Version: v1.x
* Namecheap provider version: v2.1.1

Example
```hcl
resource "namecheap_domain_records" "my-domain-com" {
  # ...
  email_type = "FWD"

  email_redirect {
    from = "*"
    to   = "support@example.com"
  }
}
```

kurok:
To manage email redirect rules using Terraform, especially for Namecheap, you need to rely on the provider's support for such features. Now, direct support for email forwarding or redirect rules through the Terraform Namecheap provider is not available out of the box. We will consider this feature request.

PR #289, opened 2026-07-10
feat: add namecheap_email_forwarding resource

kurok:
## Summary

Wraps `namecheap.domains.dns.getEmailForwarding`/`setEmailForwarding`, which has existed in `go-namecheap-sdk` since v2.7.0 (already pinned in `go.mod`, no SDK work needed) but was never surfaced by the provider — the community request in #85 has been open since 2024.

This closes only the email-forwarding half of #249; the nameserver half already shipped in #284 as `namecheap_personal_nameserver`.

- New resource `namecheap_email_forwarding`, following the `namecheap_personal_nameserver`/`namecheap_domain_contacts` SDKv2 pattern (schema + CRUD + `ValidateFunc`/`ValidateDiagFunc` + `diagFromClientError`, registered in `provider.go`).
- `setEmailForwarding` is a full-table replace, so this resource owns a domain's **entire** forwarding table: rules created outside Terraform surface as drift on refresh and are replaced on the next apply. `forwards` must be non-empty — destroy the resource to clear all forwarding.
- `forwards` is a `TypeMap` (alias → destination) with attribute-scoped plan-time validation per entry (lowercase local alias or `*` catch-all; destination must look like an email address).
- Apply-time warning (not a plan-time error — same SDKv2 limitation already documented for #250's OVERWRITE warnings) when the domain doesn't use Namecheap BasicDNS/FreeDNS, or its `email_type` isn't `"FWD"` — forwarding is stored either way but won't route mail until fixed. Costs exactly one extra `GetHosts` call per apply, zero extra calls during plan/refresh (mock call-count asserted).
- Read handles the domain-gone case by reusing `namecheap_domain_contacts`'s existing `isDomainGoneError` helper.
- Docs: `docs/resources/email_forwarding.md` (matching the existing resource-doc format; README/`docs/index.md` are intentionally untouched — neither #284 nor #285 touched them either, so that's not this repo's convention for new resources).

## Test plan

- [x] `make build format check lint test` — unit suite green; coverage 88.7% vs 88.5% on master (no regression)
- [x] `make testacc-mock` — full mock acceptance suite green, including existing domain_records/domain_contacts/nameserver/data-source suites, unmodified
- [x] New unit coverage: `forwards` validator (empty map, invalid mailbox/destination shapes, attribute-scoped diagnostics), `forwardsMapToSlice`/`forwardsSliceToMap` (deterministic ordering, case handling, dup-key last-wins), import, full CRUD incl. both conflict-warning paths and the domain-gone/API-error paths
- [x] New mock acceptance coverage: lifecycle (create/update add+change+remove/import/destroy), full-ownership + drift, call-count assertions (exactly one `getEmailForwarding` per refresh, one `setEmailForwarding` + one `getHosts` per apply, zero added calls on plan), apply succeeds despite a DNS-mode/`email_type` conflict, negative `setEmailForwarding` API-error path
- [x] Live sandbox test added (`namecheap_email_forwarding_live_test.go`, gated on `NAMECHEAP_TEST_DOMAIN`) — captures and restores the shared sandbox domain's existing forwarding table around the test, mirroring `namecheap_domain_contacts_live_test.go`'s pattern, since destroy here is a real API call (not state-only). This is the mechanism that verifies the plan's two sandbox-only assumptions (mailbox alias format is the local part; an empty `setEmailForwarding` call clears the table) — will run automatically in the live-sandbox CI job on this PR.

### DIFF
--- a/docs/resources/email_forwarding.md
+++ b/docs/resources/email_forwarding.md
@@ -1,0 +1,49 @@
+---
+page_title: "namecheap_email_forwarding Resource - terraform-provider-namecheap"
+subcategory: ""
+description: |-
+  Manages a domain's entire email forwarding table (mailbox alias -> destination address).
+---
+
+# namecheap_email_forwarding (Resource)
+
+Manages a domain's email forwarding rules via the Namecheap
+[`domains.dns.getEmailForwarding` / `domains.dns.setEmailForwarding`](https://www.namecheap.com/support/api/methods/domains-dns/set-email-forwarding/)
+API.
+
+~> **Full-ownership resource:** `setEmailForwarding` replaces the domain's **entire** forwarding table in one call, so this resource owns every rule for the domain. Forwarding rules created outside Terraform (e.g. through the dashboard) surface as drift on the next refresh and are **replaced** on the next apply. Destroying this resource clears the table entirely.
+
+~> **Requires Namecheap BasicDNS/FreeDNS and `email_type = "FWD"`:** email forwarding only takes effect when the domain uses Namecheap's default DNS and its [`namecheap_domain_records`](./domain_records.md) resource (or the dashboard) has `email_type` set to `"FWD"`. If either condition isn't met, `apply` still succeeds and stores the rules, but emits a warning — the rules will not route mail until the mismatch is fixed.
+
+## Example Usage
+
+```terraform
+resource "namecheap_domain_records" "example-com" {
+  domain     = "example.com"
+  email_type = "FWD"
+}
+
+resource "namecheap_email_forwarding" "example-com" {
+  domain = "example.com"
+
+  forwards = {
+    info  = "me@example.com"
+    sales = "sales-team@example.com"
+    # "*" is the catch-all alias
+    "*" = "catchall@example.com"
+  }
+}
+```
+
+## Argument Reference
+
+- `domain` - (Required, Force New) The registered root domain whose email forwarding is managed (e.g. `example.com`). Must be a root domain, not a subdomain. Changing this forces a new resource.
+- `forwards` - (Required) Map of mailbox alias to destination email address. Must be non-empty — destroy the resource instead of emptying this map to remove all forwarding. Each key must be a lowercase local alias with no `@` or whitespace (e.g. `info`), or `*` for a catch-all; each value must look like an email address.
+
+## Import
+
+Email forwarding can be imported by domain name, e.g.,
+
+```terraform
+terraform import namecheap_email_forwarding.example-com example.com
+```
--- a/docs/resources/email_forwarding.md
+++ b/docs/resources/email_forwarding.md
@@ -17,10 +17,19 @@ API.
 
 ## Example Usage
 
+~> `namecheap_domain_records` only sends `email_type` to the API as a side effect of writing at least one `record` or `nameservers` entry — a `namecheap_domain_records` resource with `email_type` set but no `record`/`nameservers` sends nothing to the API on create, so the domain's real `email_type` would silently stay unchanged. Include at least one `record` block, as below, so `email_type = "FWD"` actually reaches the domain.
+
 ```terraform
 resource "namecheap_domain_records" "example-com" {
   domain     = "example.com"
+  mode       = "OVERWRITE"
   email_type = "FWD"
+
+  record {
+    hostname = "@"
+    type     = "A"
+    address  = "203.0.113.10"
+  }
 }
 
 resource "namecheap_email_forwarding" "example-com" {
@@ -32,6 +41,8 @@ resource "namecheap_email_forwarding" "example-com" {
     # "*" is the catch-all alias
     "*" = "catchall@example.com"
   }
+
+  depends_on = [namecheap_domain_records.example-com]
 }
 ```
 
--- a/namecheap/mock_email_forwarding_test.go
+++ b/namecheap/mock_email_forwarding_test.go
@@ -3,15 +3,29 @@
 package namecheap_provider
 
 import (
+	"context"
 	"fmt"
 	"regexp"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 )
 
 const mockEmailForwardingDomain = "mock-forwarding-example.com"
+
+// planCheckFunc adapts a plain func to plancheck.PlanCheck, so a call-count
+// assertion can run via ConfigPlanChecks - the mechanism that actually
+// executes for a PlanOnly step. TestStep.Check does not: it is only invoked
+// for non-PlanOnly steps (terraform-plugin-testing v1.16.0,
+// helper/resource/testing_new_config.go), so a Check on a PlanOnly step is
+// silently skipped and never asserts anything.
+type planCheckFunc func(context.Context, plancheck.CheckPlanRequest, *plancheck.CheckPlanResponse)
+
+func (f planCheckFunc) CheckPlan(ctx context.Context, req plancheck.CheckPlanRequest, resp *plancheck.CheckPlanResponse) {
+	f(ctx, req, resp)
+}
 
 // mockCheckForward asserts the mock persisted the given destination for a
 // mailbox alias on the domain.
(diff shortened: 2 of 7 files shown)


namecheap/mock_email_forwarding_test.go
```diff
@@ -0,0 +1,274 @@
...
+					}
+					return nil
+				},
+			},
+			{
+				// A plan-only step against the unchanged config must issue
+				// exactly one more getEmailForwarding (the refresh read) and
+				// no getHosts at all - the conflict check only runs on apply.
+				Config:   config,
+				PlanOnly: true,
```
dm1tr1yvovk, line 233: **This step's `Check` never executes.** In terraform-plugin-testing v1.16.0, the `Check` function is only invoked inside the `if !step.PlanOnly` block (`helper/resource/testing_new_config.go:109–263`, `Check` at line 216) — `Check` on a `PlanOnly` step is silently skipped.

I verified this empirically on this branch: planting `return fmt.Errorf("BOOM")` at the top of this func still passes the test.

Worse, the assertion is wrong as written: dumping the mock's real counters shows `getEmailForwarding = 2` by this point, not 1 — SDKv2 automatically calls `ReadContext` after `CreateContext` during the step-1 apply, plus the framework's permadiff refresh. So if this check ever *did* run, it would fail.

(The `getHosts` half of the claim is true in practice — the conflict check really does run exactly once, on apply — it's just not being asserted anywhere that executes.)

Suggested fix: snapshot the counters in this step's `PreConfig` and assert the *delta* in a follow-up non-`PlanOnly` step (or use `ConfigPlanChecks`), and correct the expected numbers to account for the post-create Read.

docs/resources/email_forwarding.md
```diff
@@ -0,0 +1,49 @@
...
+API.
+
+~> **Full-ownership resource:** `setEmailForwarding` replaces the domain's **entire** forwarding table in one call, so this resource owns every rule for the domain. Forwarding rules created outside Terraform (e.g. through the dashboard) surface as drift on the next refresh and are **replaced** on the next apply. Destroying this resource clears the table entirely.
+
+~> **Requires Namecheap BasicDNS/FreeDNS and `email_type = "FWD"`:** email forwarding only takes effect when the domain uses Namecheap's default DNS and its [`namecheap_domain_records`](./domain_records.md) resource (or the dashboard) has `email_type` set to `"FWD"`. If either condition isn't met, `apply` still succeeds and stores the rules, but emits a warning — the rules will not route mail until the mismatch is fixed.
+
+## Example Usage
+
+```terraform
+resource "namecheap_domain_records" "example-com" {
```
dm1tr1yvovk, line 23: **This example silently does nothing server-side.** `resourceRecordCreate` (`namecheap/namecheap_domain_record.go:134-192`) only sends `email_type` to the API as a side effect of a records/nameservers `SetHosts` call — with zero `record` blocks and zero `nameservers`, all four branches are skipped and `email_type` is dropped (Terraform state will even claim `FWD`).

This is exactly the pitfall the live-test comment in this PR documents and works around with a throwaway TXT record — but the user-facing example ships the broken shape. A user copy-pasting it gets non-routing forwarding with no error.

Please add at least one `record` block to the example (as the live test does) and note the requirement. Since `docs/` is registry-published, better to fix inside this PR than in a follow-up (which would need to be `fix(docs):` to actually ship).

namecheap/mock_email_forwarding_test.go
```diff
@@ -0,0 +1,274 @@
...
+					mockCheckForwardCount(m, mockEmailForwardingDomain, 1),
+					mockCheckForward(m, mockEmailForwardingDomain, "info", "info-dest@example.com"),
+				),
+			},
+			{
+				// An out-of-band rule appears after Terraform has taken
+				// ownership - the next refresh must surface it as drift.
+				PreConfig: func() {
+					st := m.state(mockEmailForwardingDomain)
+					st.forwards["extra"] = "extra-dest@example.com"
```
dm1tr1yvovk: nit: this mutates the pointer returned by `m.state()` directly, violating its documented contract ("Callers must not mutate the returned pointer") and bypassing the mutex. Harmless today since no requests are in flight during `PreConfig`, but a small locked mutator (e.g. `m.addForward(domain, mailbox, dest)`) would keep the contract honest.

namecheap/mock_server_test.go
```diff
@@ -158,6 +176,16 @@ func (m *namecheapMock) seed(domain string, hosts []hostEntry, emailType string,
...
 }
 
+// seedForwards sets the initial email forwarding table for a domain,
+// simulating rules that already exist before Terraform manages them. Call it
+// before the provider issues any request (e.g. in a PreConfig/PreCheck).
+func (m *namecheapMock) seedForwards(domain string, forwards map[string]string) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	st := m.stateFor(domain)
+	st.forwards = forwards
```
dm1tr1yvovk: nit: this stores the caller's map by reference rather than copying, so later mock-side mutations alias the test's map (and vice versa). Copying the map here would match the defensive posture the `state()` contract implies.

dm1tr1yvovk (changes requested):
Solid resource implementation — I checked out the branch locally, verified the mock against SDK v2.7.0 source (`mailbox1`/`ForwardTo1` param casing and XML attr casing match exactly; API errors surface centrally via `parseAPIError`, so the error paths are sound), confirmed empirically that `forwards = {}` is rejected at plan time through real Terraform, and ran build/vet/lint/unit/mock-acceptance — all green.

Requesting changes for two contained issues (details inline):

1. **`TestAccMockEmailForwardingCallCounts` step 2 assertions are dead code** — `Check` never runs on `PlanOnly` steps, and the assertion would fail if it did run. The "zero extra API calls during plan" AC is currently unverified.
2. **The doc example's `namecheap_domain_records` block silently drops `email_type`** — it reproduces exactly the pitfall this PR's own live-test comment documents and works around. `docs/` is registry-published, so this should be fixed before merge rather than in a follow-up.

Plus two test-hygiene nits, also inline.

One residual-risk note, no action needed: the test plan says the live test verifies "an empty `setEmailForwarding` call clears the table", but given the documented sandbox read-back gap, only *destroy succeeds without error* is actually verifiable there — clearing itself can't be confirmed until the sandbox returns stored forwards. Worth keeping in mind if a real-API regression ever comes in.

namecheap/mock_email_forwarding_test.go
kurok: Confirmed empirically: forced a failure in the check body and the test still passed pre-fix. Replaced Check with ConfigPlanChecks.PostApplyPostRefresh (runs regardless of PlanOnly, confirmed against helper/resource/testing_new_config.go), asserting the delta from a PreConfig-snapshotted baseline. Re-verified with the same forced-failure technique: it now fails correctly. Fixed in d361e36.

docs/resources/email_forwarding.md
kurok: Added a record block to the namecheap_domain_records example and an explicit callout above it explaining the email_type-without-record-blocks pitfall. Fixed in d361e36.

namecheap/mock_email_forwarding_test.go
kurok: Added namecheapMock.addForward, a locked mutator, and switched this PreConfig to use it instead of mutating the state() pointer directly. Fixed in d361e36.

namecheap/mock_server_test.go
kurok: seedForwards now copies the map into a fresh map before storing it. Fixed in d361e36.

dm1tr1yvovk (dismissed):
All four review findings are correctly addressed in d361e36 — re-verified locally on the updated branch, not just by reading the replies:

- **Call-count test:** re-ran the planted-failure proof against the new `ConfigPlanChecks.PostApplyPostRefresh` check — it now genuinely executes on the PlanOnly step (forced failure fails the test; real delta assertions pass: +1 `getEmailForwarding` from the refresh, +0 `getHosts`). The delta-from-PreConfig-baseline approach also correctly accounts for SDKv2's automatic post-Create Read.
- **Doc example:** now includes a `record` block plus an explicit callout about the `email_type`-without-records pitfall, and a `depends_on` for ordering.
- **Mock hygiene:** `addForward` is a proper locked mutator (nil-safe), and `seedForwards` copies the caller's map.

Build/vet/lint (0 issues), unit tests, and the full mock acceptance suite (24/24) all green locally; all CI checks including the live acceptance test pass. Good to go. :shipit:

dm1tr1yvovk (approved):
Re-approving after the rebase onto master (previous approval was auto-dismissed by the force-push).

The rebase had a single conflict, in `mock_server_test.go`: #288 independently added the same `commandCounts`/`commandCount()` call-count infrastructure this PR introduced. Resolved by keeping master's copy — the rebased branch has exactly one accessor and one handler increment, no duplication. No functional changes to the resource itself.

Verified locally on the rebased head (c10575d): build, vet, golangci-lint (0 issues), unit tests, and the full mock acceptance suite all green; `namecheap_email_forwarding` correctly registered alongside the new #286 data sources; DCO sign-offs preserved on all 7 commits. :shipit:

Issue #256, opened 2026-07-02
Docs & registry overhaul: tfplugindocs generation, validated examples tree, CI/import/portfolio guides


kurok:
## Problem

Provider docs are hand-maintained markdown (`docs/index.md`, one resource page, two guides) — no `tfplugindocs` generation, no `examples/` tree, and no template structure. The failure mode is already visible in the issue history: recurring support questions answered in issues instead of docs (#67/#62/#59 CI IP whitelisting asked three times; #63 import confusion; #35 nameserver semantics). With this roadmap adding ~10 resources/data sources, hand-maintained docs will fall behind by the second release — schema drift between code and docs is guaranteed unless generation is wired in first.

### 1. tfplugindocs generation

- Adopt `tfplugindocs`: schema descriptions live in code (framework `MarkdownDescription` on every attribute), templates in `templates/`, generated output in `docs/` (registry-published; the repo's CLAUDE.md already reserves `docs/` for registry content).
- CI check: `make docs` regenerates; a diff-check job fails PRs whose docs are stale (same pattern as the vendor-consistency check from #169).
- Existing pages (index, domain_records, the two guides) migrated into templates so nothing published today regresses.

### 2. `examples/` tree (feeds generation + humans)

- `examples/provider/`, `examples/resources/<name>/resource.tf`, `examples/data-sources/<name>/data-source.tf` — the layout tfplugindocs pulls into pages automatically; every example is `terraform validate`-able (CI job validates all of them against the built provider — examples that rot are worse than none).

### 3. Guides for the recurring questions

- **"Running in CI"**: IP whitelisting for dynamic runners, `client_ip` auto-detection, rate-limit tuning (canonicalizes #67/#62/#59 — three issues become one link).
- **"Importing an existing portfolio"**: import blocks, per-resource ID formats, domain_records vs dns_record choice (#63/#68 class).
- **Registry front page** (`docs/index.md`): full provider config reference (generated), env-var table, sandbox usage, supported TF/OpenTofu versions.
- Existing v2.0.0 migration guide preserved; 3.0.0 guide slot prepared (content owned by the framework-migration and OVERWRITE-safety issues).

## Acceptance criteria

- [ ] `make docs` generates the full `docs/` tree from schema + templates; CI stale-docs check fails on drift (tested by intentionally breaking once in the PR).
- [ ] Every existing resource/data source at time of merge has: generated arguments/attributes reference from in-code descriptions, at least one validated example, and an import section where applicable.
[...]

PR #138, opened 2026-04-14
chore: add git privacy rule to CLAUDE.md

kurok:
## Summary
- Add a "Git privacy" section to CLAUDE.md instructing Claude Code to use GitHub noreply email addresses when committing
- Prevents accidental exposure of contributors' real email addresses in git history

## Test plan
- [x] Verify CLAUDE.md renders correctly on GitHub
- [x] Confirm Claude Code picks up the new rule in future sessions

PR #310, opened 2026-08-01
feat: add namecheap_dns_record for per-record DNS management

### LINKED ISSUE #248: New resource: namecheap_dns_record — per-record DNS management with conflict-safe partial updates
## Problem

`namecheap_domain_records` is a **monolithic** resource: one instance owns the *entire* record set of a domain (MERGE mode softens but doesn't solve it). This shape causes the provider's most persistent pain:

- Two modules/teams/workspaces cannot each manage their own records on a shared domain without trampling each other;
- OVERWRITE mode silently deletes records created outside Terraform — open issue #65 is exactly this data-loss report;
- Common patterns from every mainstream DNS provider (`aws_route53_record`, `cloudflare_record`, `google_dns_record_set`) — for_each over records, per-record import, per-record lifecycle — don't translate.

The blocker was always the API: `setHosts` replaces everything, so per-record semantics need conflict-safe read-modify-write. That lands SDK-side as go-namecheap-sdk#119 (`AddRecords`/`DeleteRecords`/`UpsertRecords` + `ErrConcurrentModification` + `Plan` diff). This issue builds the resource on top.

### 1. `namecheap_dns_record` — one resource, one record

```hcl
resource "namecheap_dns_record" "blog" {
  domain   = "example.com"
  hostname = "blog"
  type     = "A"            # A|AAAA|ALIAS|CAA|CNAME|MX|MXE|NS|TXT|URL|URL301|FRAME
  address  = "10.11.12.13"
  ttl      = 1799           # optional, API default
  mx_pref  = 10             # MX only; validated per type at plan time
}
```

- Framework-native (per the migration issue), plan-time per-type validation (MX requires mx_pref; CAA value shape; URL/FRAME target syntax — encoding the #37/#73 lessons as validators).
- CRUD via the SDK's record-level ops: each operation touches **only its own record**, preserving everything else; `ErrConcurrentModification` retried via the SDK's `RetryOnConflict`, surfacing a clear diagnostic if contention persists.
- **Concurrency**: parallel applies on the same domain (Terraform default parallelism × for_each) are safe — the existing per-domain `mutexkv` serializes within one provider instance, and conflict detection covers cross-process races (both paths acceptance-tested).
- Import: `terraform import namecheap_dns_record.blog "example.com/A/blog/10.11.12.13"` (documented ID format; ambiguous matches → error listing candidates).

### 2. Coexistence and migration

- `namecheap_dns_record` and `namecheap_domain_records` (monolithic) are **mutually exclusive per domain** — documented loudly, and the monolithic resource's docs updated to say which to choose (new configs → per-record).
- Migration guide: from `namecheap_domain_records` to per-record via import blocks (TF 1.5+ `import {}` + a documented one-liner to generate them from current state), with a worked example.
- The monolithic resource is **not** deprecated in this issue (zone-file-style full ownership remains a valid pattern, and OVERWRITE gets its safety rail in a separate roadmap item).

## Acceptance criteria

- [ ] Full lifecycle matrix (per the testing-infrastructure issue) green against the mock: create/update/replace/import/destroy/drift for every record type the API supports, including MX+mx_pref, CAA iodef (#73 regression), and subdomain-vs-root placement (#71 regression).
- [ ] Surgical updates proven: a domain seeded with 10 records, managing 1 via the resource — create/update/destroy leaves the other 9 byte-identical (mock request assertions on the setHosts payload).
- [ ] Two records on the same domain applied with parallelism=10 succeed without lost updates (acceptance test); simulated cross-process conflict (scripted mock) → retry → success, and persistent conflict → actionable diagnostic.
- [ ] Plan-time validators reject invalid per-type combinations with attribute-scoped errors (table-driven).
- [ ] Import round-trips every documented ID form; ambiguity error lists candidates (tested).
- [ ] Mutual-exclusion guidance + migration guide (with import-block example) published in `docs/`; `namecheap_domain_records` docs cross-reference.
- [ ] Registry docs for the resource generated (tfplugindocs); examples/ entry added.
- [ ] `make build format check lint test testacc` green; conventional commits (`feat:` → minor release via release-please).

## Out of scope

- Deprecating `namecheap_domain_records` (future decision, data-driven).
- DNSSEC (#83 — blocked on API surface investigation, tracked in the epic).
- Email-forwarding records (separate quick-win resource issue).

Depends on: go-namecheap-sdk#119 (record-level ops + conflict detection), framework migration phase A/B, testing infrastructure. Fixes the workflow behind #65. Part of the provider 3.0 roadmap (see tracking issue).

kurok:
Closes #248.

`namecheap_domain_records` owns a domain entire zone, which forces every record into one resource and one owner. Two teams cannot each manage their own records, iterating over a map of hosts is impossible, and a record created outside Terraform is either adopted wholesale or deleted. This adds the shape people expect from a DNS provider: one resource per record, per-record import, independent instances sharing a zone.

## How a change is applied

Namecheap has no per-record API — setHosts replaces the whole record set — so every change is a read-modify-write of the entire zone, guarded two ways. Within a run, changes to one domain are serialized through the existing per-domain mutex. Across runs, the SDK record layer re-reads the zone after writing and retries when another writer changed it, so a lost update surfaces as an error rather than a silently missing record. That layer is what go-namecheap-sdk#119 and #138 were built for, and this is its first consumer.

## Design decisions worth reviewing

**Identity versus attributes.** domain, hostname and type force replacement; address, ttl and mx_pref change in place. An in-place update of a hostname would rewrite a different record, so this is a correctness boundary rather than a preference, and a test pins it.

**Lookups match on identity alone.** My first version compared whole records including TTL. That broke import, where TTL is unknown until the record is found, and would have made a TTL changed in the dashboard look like a deletion. Host, type and address identify a record; TTL is an attribute of it.

**Creating a record that already exists is refused**, with the import command that adopts it — setHosts would otherwise write a second copy that no selector could tell apart afterwards.

**A record deleted outside Terraform drops out of state** so the next plan recreates it.

## Compatibility

Purely additive: a new resource, registered alongside the others. No existing resource, data source, schema or default is touched, so nothing in anyone state can move. The SDK bump to v2.10.0 brings the EmailType work from go-namecheap-sdk#161, which is additive there too.

## Known limitation, documented rather than hidden

Domains on Private Email or Google Workspace **cannot be managed with this resource at all** — not even to change an unrelated A record. Those zones hold their provider MX records while the zone type stays OX or GMAIL, a combination the SDK setHosts validation rejects. Whether that validation is even correct is unresolved: those rules appear nowhere in the published API reference, and the API itself returns a record set the SDK then refuses to write back. Filed as go-namecheap-sdk#162, which needs one sandbox round-trip against a real Private Email domain to settle. The resource page says so plainly and points those users at namecheap_domain_records.

## Tests

8 unit tests: ID rendering, the import round trip with slashes inside the address, identity-versus-attribute selection, a guard that a selector can never degenerate into matching every record, and the schema contract.

6 mock acceptance tests through real Terraform: create, in-place update, identity replacement, several records managed independently on one domain, duplicate refusal, drift recreation, and three import formats including a malformed one. **Every one of them seeds records nobody manages and asserts they survive** — surgical modification is the entire point of the resource, so it is what the suite is built around.

The docs job added in #256 caught the missing page on its first real use, which is a pleasant confirmation that it works. Locally green: build, vet, lint, unit suite, the full 47-test mock acceptance suite, docs-check and examples-validate (17 examples).

### DIFF
--- a/docs/resources/domain_host_record.md
+++ b/docs/resources/domain_host_record.md
@@ -118,19 +118,26 @@ Tracked in [go-namecheap-sdk#162](https://github.com/namecheap/go-namecheap-sdk/
 - `domain` - (Required, Force New) The registered root domain the record belongs to (e.g. `example.com`). Must be a root domain on the account, not a subdomain.
 - `hostname` - (Required, Force New) The sub-domain the record answers for, or `@` for the domain itself (e.g. `www`). Must not be empty — write `@` for the apex.
 - `type` - (Required, Force New) The record type: `A`, `AAAA`, `ALIAS`, `CAA`, `CNAME`, `MX`, `MXE`, `NS`, `TXT`, `URL`, `URL301`, `FRAME`.
-- `address` - (Required) The record's value, whose meaning depends on `type`: an IP address for `A`/`AAAA`, a hostname for `CNAME`/`MX`/`NS`, arbitrary text for `TXT`, a URL for `URL`/`URL301`/`FRAME`. Changed in place.
-- `ttl` - (Optional) Time to live in seconds, between `60` and `60000`. Defaults to `1800`. Changed in place.
-- `mx_pref` - (Optional) MX preference, lower being preferred, between `0` and `255`. Defaults to `10`. Changed in place.
+- `address` - (Required) The record's value, whose meaning depends on `type`: an IP address for `A`/`AAAA`, a hostname for `CNAME`/`MX`/`NS`, arbitrary text for `TXT`, a URL for `URL`/`URL301`/`FRAME`. Edited on the existing record; not Force New.
+- `ttl` - (Optional) Time to live in seconds, between `60` and `60000`. Defaults to `1800`. Edited on the existing record; not Force New.
+- `mx_pref` - (Optional) MX preference, lower being preferred, between `0` and `255`. Defaults to `10`. Edited on the existing record; not Force New.
 
 -> `mx_pref` applies to `MX` records only, where it is part of the record's
 identity — a primary and a backup mail server may name the same host, and the
 preference is what tells them apart. For every other type Namecheap stores a fixed
 `10` whatever is sent, so a value set there is ignored.
 
-`domain`, `hostname` and `type` are the record's identity, so changing any of
-them replaces the record rather than editing it. `address`, `ttl` and `mx_pref`
-are changed in place; if such a change is rejected by the API, the record is left
-as it was and the next plan proposes the same change again.
+### What replaces a record and what edits it
+
+`domain`, `hostname` and `type` are the record's identity, so changing any of them
+**destroys the record and creates a new one** — the plan says
+`-/+ destroy and then create replacement`.
+
+`address`, `ttl` and `mx_pref` are **edited on the record that is already there**:
+the plan shows `~ update in-place`, one write swaps the value, and the name keeps
+resolving throughout. Changing `address` also re-renders the resource's `id`, which
+contains it. If such a change is rejected by the API, the record is left exactly as
+it was and the next plan proposes the same change again.
 
 -> The case of `domain` and `hostname` is kept as you write it, as is an address
 the API only re-spells (the trailing dot it adds to a `CNAME`, `ALIAS`, `NS` or
--- a/docs/resources/domain_host_record.md
+++ b/docs/resources/domain_host_record.md
@@ -1,11 +1,11 @@
 ---
-page_title: "namecheap_dns_record Resource - terraform-provider-namecheap"
+page_title: "namecheap_domain_host_record Resource - terraform-provider-namecheap"
 subcategory: "DNS"
 description: |-
-  {{ .Description }}
+  Manages a single DNS host record on a domain, leaving all other records untouched. Mutually exclusive with namecheap_domain_records for the same domain.
 ---
 
-# namecheap_dns_record (Resource)
+# namecheap_domain_host_record (Resource)
 
 Manages a **single** DNS host record, leaving every other record on the domain
 untouched. This is the per-record counterpart to
@@ -15,11 +15,45 @@ per record, `for_each` over a map, per-record import.
 
 ## Example Usage
 
-{{tffile "examples/resources/dns_record/example_1.tf"}}
+```terraform
+# Each record is managed independently; everything else in the zone is left alone.
+resource "namecheap_domain_host_record" "www" {
+  domain   = "example.com"
+  hostname = "www"
+  type     = "A"
+  address  = "203.0.113.10"
+  ttl      = 1800
+}
+
+resource "namecheap_domain_host_record" "spf" {
+  domain   = "example.com"
+  hostname = "@"
+  type     = "TXT"
+  address  = "v=spf1 include:_spf.example.net -all"
+}
+```
 
 ### One record per entry
 
-{{tffile "examples/resources/dns_record/example_2.tf"}}
+```terraform
+# A record per entry, driven by a map. Each instance owns exactly its own record,
+# so adding or removing one key does not touch the others.
+locals {
+  hosts = {
+    api = "203.0.113.20"
+    cdn = "203.0.113.21"
+  }
+}
+
+resource "namecheap_domain_host_record" "fleet" {
+  for_each = local.hosts
+
+  domain   = "example.com"
+  hostname = each.key
+  type     = "A"
+  address  = each.value
+}
+```
 
 ## Choosing between this and `namecheap_domain_records`
 
@@ -29,28 +63,21 @@ together on one domain they fight — the zone-level resource will delete record
 this one created (in `OVERWRITE` mode) or fight over ownership (in `MERGE`),
 because neither knows about the other.
 
-- **Use `namecheap_dns_record`** when different records are owned by different
-  people, modules or workspaces, or when you want `for_each` over a map of hosts.
+- **Use `namecheap_domain_host_record`** when different records are owned by
+  different people, modules or workspaces, or when you want `for_each` over a map
+  of hosts.
 - **Use `namecheap_domain_records`** when one configuration should own the whole
   zone, zone-file style — including the guarantee that nothing else exists.
 
-## How a change is applied
+## Concurrent changes to one domain
 
-Namecheap has no per-record API: `domains.dns.setHosts` replaces the entire
-record set. Every change here is therefore a read-modify-write of the whole zone,
-protected two ways:
-
-1. Within a single Terraform run, changes to one domain are serialized, so
-   parallel instances of this resource do not interleave.
-2. After writing, the zone is re-read and compared against what the write
-   intended. A mismatch is retried, so a change of *yours* is not silently lost.
-
-~> **This narrows the race with writers outside the run; it does not close it.**
-The comparison is against the record set this run computed from its own read, so a
-foreign write that lands between that read and the `setHosts` is inside the
-replaced set — it is overwritten, the verification still matches, and the apply
-succeeds. Concurrent writers *can* cost you a record. Serialize changes to a
-domain: one Terraform run at a time, and no dashboard edits while it runs.
+!> **Change a domain from one place at a time.** Namecheap has no per-record API,
+so a change to any single record rewrites the domain's whole record set. Records
+managed within one Terraform run are safe — the provider serializes them, and it
+detects and retries most interference from outside the run — but that window
+cannot be closed: a second writer landing mid-apply, another `terraform apply` or
+an edit in the Namecheap dashboard, can still cost you a record silently. Run one
+apply per domain at a time, and leave the zone alone while it runs.
 
 ## Records this resource cannot tell apart
 
@@ -98,9 +125,7 @@ Tracked in [go-namecheap-sdk#162](https://github.com/namecheap/go-namecheap-sdk/
 -> `mx_pref` applies to `MX` records only, where it is part of the record's
 identity — a primary and a backup mail server may name the same host, and the
 preference is what tells them apart. For every other type Namecheap stores a fixed
-`10` whatever is sent, so a value set there is ignored rather than fought over:
-sending it would make the post-write check compare it against the API's `10` and
-report the write as a lost race.
+`10` whatever is sent, so a value set there is ignored.
 
 `domain`, `hostname` and `type` are the record's identity, so changing any of
 them replaces the record rather than editing it. `address`, `ttl` and `mx_pref`
@@ -119,7 +144,11 @@ the API only re-spells (the trailing dot it adds to a `CNAME`, `ALIAS`, `NS` or
 
 Records are imported by that same composite ID:
 
-{{codefile "shell" "examples/resources/dns_record/import.sh"}}
+```shell
+# The ID is <domain>/<type>/<hostname>/<address>. The address may itself contain
+# slashes (a URL record, for instance) — only the first three components are split.
+terraform import namecheap_domain_host_record.www example.com/A/www/203.0.113.10
+```
 
 The address may itself contain `/` — a `URL` record's target, for instance —
 because only the first three components are split off.
--- a/namecheap/namecheap_domain_host_record.go
+++ b/namecheap/namecheap_domain_host_record.go
@@ -13,47 +13,55 @@ import (
 )
 
 const (
-	// dnsRecordIDSeparator joins the four components of a namecheap_dns_record
+	// hostRecordIDSeparator joins the four components of a namecheap_domain_host_record
 	// import ID: domain/type/hostname/address.
-	dnsRecordIDSeparator = "/"
+	hostRecordIDSeparator = "/"
 
-	// dnsRecordRetryAttempts is how many times a mutation re-runs the SDK's
+	// hostRecordRetryAttempts is how many times a mutation re-runs the SDK's
 	// read-modify-write-verify cycle when it loses a race with another writer.
 	// Namecheap has no per-record API — every change rewrites the whole zone —
 	// so two applies touching one domain concurrently will collide, and the
 	// retry is what makes for_each over a domain's records usable at all.
-	dnsRecordRetryAttempts = 3
+	hostRecordRetryAttempts = 3
 
-	// dnsRecordFixedMXPref is the preference Namecheap reports for every record
+	// hostRecordFixedMXPref is the preference Namecheap reports for every record
 	// type that does not have one — which is all of them but MX. It is both the
 	// schema default and the value sent for those types; see
-	// dnsRecordEffectiveMXPref for why sending anything else cannot work.
-	dnsRecordFixedMXPref = 10
+	// hostRecordEffectiveMXPref for why sending anything else cannot work.
+	hostRecordFixedMXPref = 10
 )
 
-// resourceNamecheapDNSRecord manages a single DNS host record, leaving every
+// resourceNamecheapDomainHostRecord manages a single DNS host record, leaving every
 // other record on the domain untouched.
 //
 // This is the per-record counterpart to namecheap_domain_records, which owns a
 // domain's whole zone. The two are mutually exclusive per domain: pointing both
 // at the same domain makes each fight to impose its own view of the zone.
 //
+// How a change is applied — the mechanics the resource page deliberately does not
+// carry, because none of it is actionable for a user beyond "change a domain from
+// one place at a time":
+//
 // The underlying API has no per-record operation — domains.dns.setHosts replaces
-// the entire record set — so every change here is a read-modify-write of the
-// whole zone, guarded two ways: the provider serializes changes to one domain
-// through ncMutexKV within a single Terraform run, and the SDK verifies the zone
-// after writing so a change lost to a writer outside this run surfaces as an
-// error instead of silently disappearing.
-func resourceNamecheapDNSRecord() *schema.Resource {
+// the entire record set — so every change here is a read-modify-write of the whole
+// zone, guarded two ways. The provider serializes changes to one domain through
+// ncMutexKV within a single Terraform run. Across runs, the SDK re-reads the zone
+// after writing and retries when the result is not what it wrote, which catches a
+// concurrent writer in most cases but cannot catch all of them: the comparison is
+// against the set computed from this run's own read, so a foreign write landing
+// between that read and the setHosts is inside the replaced set — overwritten,
+// verified as correct, and lost. Narrowed, not closed; see the warning on the
+// resource page.
+func resourceNamecheapDomainHostRecord() *schema.Resource {
 	return &schema.Resource{
 		Description: "Manages a single DNS host record on a domain, leaving all other records untouched. Mutually exclusive with namecheap_domain_records for the same domain.",
 
-		CreateContext: resourceNamecheapDNSRecordCreate,
-		ReadContext:   resourceNamecheapDNSRecordRead,
-		UpdateContext: resourceNamecheapDNSRecordUpdate,
-		DeleteContext: resourceNamecheapDNSRecordDelete,
+		CreateContext: resourceNamecheapDomainHostRecordCreate,
+		ReadContext:   resourceNamecheapDomainHostRecordRead,
+		UpdateContext: resourceNamecheapDomainHostRecordUpdate,
+		DeleteContext: resourceNamecheapDomainHostRecordDelete,
 		Importer: &schema.ResourceImporter{
-			StateContext: resourceNamecheapDNSRecordImport,
+			StateContext: resourceNamecheapDomainHostRecordImport,
 		},
 
 		Schema: map[string]*schema.Schema{
@@ -91,10 +99,10 @@ func resourceNamecheapDNSRecord() *schema.Resource {
 			"mx_pref": {
 				Type:         schema.TypeInt,
 				Optional:     true,
-				Default:      dnsRecordFixedMXPref,
+				Default:      hostRecordFixedMXPref,
 				ValidateFunc: validation.IntBetween(0, 255),
 				Description: fmt.Sprintf("The MX preference, lower being preferred, between 0 and 255. Applies to MX records only — Namecheap stores a fixed %d for every other type, so the value is ignored there rather than fought over. "+
-					"For MX records it is part of what identifies the record, so a primary and a backup mail server can name the same host.", dnsRecordFixedMXPref),
+					"For MX records it is part of what identifies the record, so a primary and a backup mail server can name the same host.", hostRecordFixedMXPref),
 			},
 			"ttl": {
 				Type:         schema.TypeInt,
@@ -107,19 +115,19 @@ func resourceNamecheapDNSRecord() *schema.Resource {
 	}
 }
 
-// dnsRecordFromData builds the SDK record described by the configuration.
-func dnsRecordFromData(data *schema.ResourceData) namecheap.DomainsDNSHostRecord {
+// hostRecordFromData builds the SDK record described by the configuration.
+func hostRecordFromData(data *schema.ResourceData) namecheap.DomainsDNSHostRecord {
 	recordType := data.Get("type").(string)
 	return namecheap.DomainsDNSHostRecord{
 		HostName:   namecheap.String(strings.ToLower(data.Get("hostname").(string))),
 		RecordType: namecheap.String(strings.ToUpper(recordType)),
 		Address:    namecheap.String(data.Get("address").(string)),
-		MXPref:     namecheap.UInt8(dnsRecordEffectiveMXPref(recordType, data.Get("mx_pref").(int))),
+		MXPref:     namecheap.UInt8(hostRecordEffectiveMXPref(recordType, data.Get("mx_pref").(int))),
 		TTL:        namecheap.Int(data.Get("ttl").(int)),
 	}
 }
 
-// dnsRecordEffectiveMXPref returns the preference to send for a record of this
+// hostRecordEffectiveMXPref returns the preference to send for a record of this
 // type. Only MX has one; for every other type Namecheap stores a fixed 10 whatever
 // it is sent, so the configured value is ignored rather than fought over.
 //
@@ -128,18 +136,18 @@ func dnsRecordFromData(data *schema.ResourceData) namecheap.DomainsDNSHostRecord
 // preference is part of that comparison. A 20 sent for an A record comes back as
 // 10, the write looks like a race it lost, and the apply fails with a concurrent
 // modification error after three pointless rewrites of the whole zone.
-func dnsRecordEffectiveMXPref(recordType string, mxPref int) uint8 {
-	if dnsRecordMXPrefIsIdentity(recordType) {
+func hostRecordEffectiveMXPref(recordType string, mxPref int) uint8 {
+	if hostRecordMXPrefIsIdentity(recordType) {
 		return uint8(mxPref)
 	}
-	return dnsRecordFixedMXPref
+	return hostRecordFixedMXPref
 }
 
-// dnsRecordBeforeChange builds the SDK record as the zone still holds it: the
+// hostRecordBeforeChange builds the SDK record as the zone still holds it: the
 // pre-change value of every attribute an update is able to move. That is what
 // identifies the record during an update, both for the ambiguity check and for
 // the selector — the new values are not in the zone yet.
-func dnsRecordBeforeChange(data *schema.ResourceData) namecheap.DomainsDNSHostRecord {
+func hostRecordBeforeChange(data *schema.ResourceData) namecheap.DomainsDNSHostRecord {
 	oldAddress, _ := data.GetChange("address")
 	oldTTL, _ := data.GetChange("ttl")
 	oldMXPref, _ := data.GetChange("mx_pref")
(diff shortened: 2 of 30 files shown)


kurok:
## Blocked — review found five reproducible defects, four of them data-loss

Recording these on the PR so they are not lost. **Do not merge this branch as it stands.** Every item below was reproduced against the mock and none is caught by the committed suite.

### D1 — Deleting one record deletes every record sharing host+type+address

The selector is built from host/type/address only, and the SDK deletes *all* matches. Seed `twin A 10.0.0.1 ttl=300` and `twin A 10.0.0.1 ttl=7200`, import one, destroy it — both are gone.

The MX variant is worse: two `@ MX mail.example.com` at prefs 10 and 20, destroy one, and the resulting set has zero MX records on an `EmailType=MX` zone, so the write is rejected with a message about email routing that points nowhere near the real cause, leaving a dangling resource. The same collision makes create refuse a legitimate second MX and hand out an ambiguous import ID.

The accepted plan required `MXPref` in the selector and an ambiguous-match error listing candidates. Neither exists; the lookup silently takes the first match.

### D2 — A failed update orphans the old record permanently

SDKv2 persists planned values when `UpdateContext` returns an error, and the update never rolls `address` back. After a failed write the zone holds both addresses, state points at the new one, and Terraform will never remove the old. It fires on exactly the failure the design says to expect (`ErrConcurrentModification`), and for an A record it means traffic keeps flowing to a decommissioned host, silently.

### D3 — Any uppercase in `domain` or `hostname` causes permanent forced replacement

Read writes the lower-cased value into state, so `domain = "Example.com"` plans `~ domain = "example.com" -> "Example.com" # forces replacement` on every run. Both fields are ForceNew, so **every apply destroys and recreates a live record, forever**. The sibling resource avoids this by lower-casing locally and never writing `domain` back.

### D4 — Permanent non-empty plan for CNAME/ALIAS/NS/MX without a trailing dot

Read stores the API's spelling, so `hosting.example.com` plans against `hosting.example.com.` on every run. Contradicts this page's own claim that the first plan after import is empty, and burns a full-zone read-modify-write each time.

### D5 — The concurrency guarantee in the docs is false

The page says a concurrent writer "cannot cost you a record silently". The SDK's verify compares the re-read against its own intended set, computed from a possibly stale read — so a foreign write landing between the SDK's read and its `setHosts` is silently deleted and the apply succeeds. The window is narrowed, not closed, and the wording has to say so.

### Also

- **TTL default should be 1799, not 1800.** The API assigns 1799 for an automatic TTL, so the first apply after any import silently pins it. The plan specified this and I asserted the wrong value as correct.
- **The mail-zone limitation is far wider than documented**: any zone holding an MX record whose EmailType is not exactly `MX` is completely unmanageable — including `FWD`, which this provider ships an email-forwarding resource for.
- **Contradictory advice**: the page tells you to set `email_type` via `namecheap_domain_records` on the same domain, four paragraphs after forbidding both resources on one domain.

### The tests are weaker than claimed

Rewriting TTL and MXPref on every preserved record passes all 7 unit tests and all 47 acceptance tests — `assertUnmanagedZoneIntact` only compares `name|type -> address`, ignoring TTL, MXPref, record count and duplicates. "Surgical modification" is the property this resource exists for and the suite does not actually assert it.

Three tests also do not test what their names say: the import round-trip exercises a hand-rolled reimplementation of the parser rather than the parser; the normalization test never calls the lookup; and the never-matches-all test asserts a field the constructor never sets, so it cannot fail.

### Minimum to unblock

1. Put `MXPref`/TTL in the selector, or refuse to act when it matches more than one record (D1).
2. Roll `address` back in state on update failure (D2).
3. Stop writing normalized `domain`/`hostname`/`address` into state — preserve the configured form when normalized-equal (D3, D4).
4. Correct the concurrency wording to what the SDK actually guarantees (D5).
5. TTL default 1799.
6. Make `assertUnmanagedZoneIntact` a full record-set comparison, and add MX and managed-CNAME coverage plus a conflict-injection test.

kurok:
All five are fixed in c923c8b. Each one got a mock acceptance test that reproduces the defect first — I watched all five fail for the reported reason before touching the resource.

**D1 — collision on host+type+address.** Two things were wrong, so there are two fixes.

`mx_pref` is now part of a record's identity for `MX` (and only `MX` — everywhere else the API reports a fixed 10). A primary and a backup naming the same host are separable again: creating the second one is no longer refused as a duplicate, and destroying either no longer takes both. That was the worst variant — on an `EmailType=MX` zone it left a record set the API refuses to write at all, so the destroy failed with a mail-routing message and a dangling resource.

Records that agree on *everything* selectable and differ only in TTL still can't be told apart — `RecordSelector` has no TTL field, so there is no way to ask the API for one of them. Those now fail with the ambiguous-match error the plan called for, listing the candidates, checked on create, read, update, delete **and** import. Previously the lookup took the first match.

```
Error: Ambiguous DNS record on example.com

  2 records on example.com share this one's identity — type A, host "twin",
  address "10.0.0.1" — and differ only in ways this resource cannot select on:

    - TTL 300
    - TTL 7200
```

**D2 — failed update orphaned the old record.** Confirmed exactly as described: after a rejected write, state held `10.6.0.2` while the zone held `10.6.0.1`, and nothing ever mentioned `10.6.0.1` again. The error path now restores the pre-change `address`, `ttl` and `mx_pref`, so the next plan proposes the same update instead. The test asserts the *original* config re-plans empty after a failed apply.

**D3 / D4 — permanent diffs from the API's spelling.** Same root cause, one fix: `Read` no longer writes `domain`, `hostname`, `type` or `address` back at all. It located the record *by* those fields (normalized), so state already described it correctly — writing the API's version was pure damage. `Import` keeps the address as the ID spelled it for the same reason; host and type are canonicalized, which is what the ID itself renders.

**D5 — the concurrency claim.** You're right, and the mechanism is as you describe: the verification compares against a set computed from this run's own read, so a foreign write inside that window is overwritten and the apply still succeeds. The page now says the race is narrowed rather than closed, and tells the reader to serialize writes to a domain.

One extra, found while adding the MX identity: `mx_pref` had no bounds, and `setHosts` takes an unsigned byte — so `256` was truncated to `0`, the *most* preferred, the opposite of the ask. Now validated at 0–255.

`make check`, `lint`, `test` and the full `testacc-mock` suite pass locally. Codecov will still flag the patch: the CRUD paths are covered only by the mock acceptance suite, and nothing uploads that suite's coverage — worth fixing separately.

dm1tr1yvovk (changes requested):
Strong PR overall — the identity-vs-attribute model, the case/trailing-dot round-trip handling, and the failed-update state restoration are all correct and well pinned by tests. Two high-level findings before merge:

**1. Update can create the duplicate that Create refuses.** Create guards against writing a record whose identity already exists, but Update only validates the *before* record and then upserts — the SDK's upsert is `filterOut(selector) + append` with no collision check. Changing `address` (or `mx_pref` on MX) onto the identity of an existing unmanaged record writes two records the selector can no longer tell apart: the post-update Read fails with the ambiguity error and the resource is permanently wedged, fixable only in the dashboard. Please add the same "already exists — import it" guard on the *target* identity in Update when an identity-relevant field changed (skip it when only `ttl` moved, since the target lookup would find the record itself).

**2. None of the new acceptance tests ran in this PR's CI.** The mock acceptance job only runs for fork/Dependabot PRs, and the sandbox job (`make testacc`) compiles without `-tags testacc`, so `mock_dns_record_test.go` was excluded entirely — the resource's CRUD paths were exercised only locally. This is also why `codecov/patch` is red.

**Ask: please add real (live sandbox) acceptance tests for `namecheap_dns_record`** — at minimum a lifecycle test (create / in-place update / import / destroy, with unmanaged records surviving) so the sandbox job actually exercises this resource against the real API on every same-repo PR. That's especially valuable here given the open question in go-namecheap-sdk#162 about whether the SDK's EmailType validation matches real API behavior; an MX case in the live suite would help settle it.

Minor, non-blocking: `hostname`/`address` accept `""` (Terraform's `Required` doesn't reject empty strings), which silently means `@` via SDK normalization but renders an ID the importer itself rejects — `validation.StringIsNotEmpty` would close that; and an explicit `mx_pref` on a non-MX record becomes a perpetual diff since Read overwrites it with the API's fixed 10.

kurok:
Both findings addressed, plus both minors. Everything green, including the live sandbox job.

## 1. Update creating the duplicate Create refuses

Confirmed and fixed in `ad5d8f2`. Reproduced first: the write landed, the post-update read returned the ambiguity error, and the resource was wedged exactly as you described.

Update now applies the same guard to the *target* identity, and only when the identity actually moved — a lone `ttl` change would otherwise find the resource's own record and call it a collision, as you noted. Both checks share one zone read, so an update costs no more API calls than before.

Your finding also exposed a second hole next to it: SDKv2 persists planned values on *any* update error, so a **refused** update was orphaning the old record just as a failed write did. Every failure in update now funnels through one state-restoring path rather than only the write error doing so.

## 2. Live acceptance coverage

Four tests in `namecheap_dns_record_live_test.go` (`f7ecfcc`, refined in `1751284` and `a6712c4`), all passing against the sandbox:

- **Lifecycle** — create, in-place update, replacement via hostname, `ImportStateVerify`, destroy, asserting against the API at every step that an unmanaged record survives.
- **RoundTrips** — proves the trailing-dot and case handling are real API behaviour rather than something the mock invented. A TXT value with spaces also round-trips verbatim.
- **MXPreference** — a managed MX beside an unmanaged one on an `EmailType=MX` zone the resource does not own; destroy leaves the primary.
- **Refusals** — create over an existing record and moving one onto another's identity, both refused before any write, then a plan proving the refusal changed nothing.

Cost is ~4 minutes of the job's 10; the file header documents the per-apply call cost and the 10/min split so the next person can size an addition.

On the CI gap you flagged: `9f8adf0` (pushed before your review landed) moved the mock suite into the `check` job, so it now runs on every PR — a same-repo PR had never executed it. That also fixed `codecov/patch`, which is green at 86%.

## 3. Your MX ask produced two answers for go-namecheap-sdk#162

This was the most valuable part of the request — the sandbox refuses two MX shapes, and the live suite now logs the evidence instead of my guess:

```
adding a second MX for "mail1.…" at preference 20 returned concurrent modification
detected; the zone then held 1 MX record(s) for that host [pref=10 ttl=1800]

changing "mail3.…" from preference 20 to 30 returned concurrent modification
detected; the zone then held [pref=20 ttl=1800]
```

So this sandbox appears to key MX records by (host, exchange) and ignore the preference on write: it will not store two MX records for one exchange, and will not change a preference in place. Both surface as a lost race, because the SDK verifies by re-reading and the zone never contains what it was sent — and the retry cannot help.

Two consequences worth recording:

- The provider is not at fault, and I did not want to take it on trust: `TestAccMockDNSRecordMXPreferenceChangesInPlace` drives exactly that preference change against a mock that is now faithful about preferences, and the record is replaced in place with the primary untouched. So the assertions stay in the mock suite and the live suite probes the shapes non-gatingly.
- I have **not** verified this against production, so I have not special-cased anything on the strength of sandbox behaviour. Keeping the preference in the record's identity is still right: without it, a zone that *does* hold such a pair loses both records to one delete. If production behaves like the sandbox, the honest follow-up is a clearer error than "re-run the apply" for a preference change that can never land — happy to open that on the SDK issue.

## Minors

- `hostname`/`address` now reject the empty string.
- `mx_pref` outside MX turned out worse than a perpetual diff. The mock was echoing the submitted preference back; making it faithful (the API reports a fixed 10) showed the apply **failing outright** with "concurrent change" after three pointless zone rewrites, because the SDK's post-write check compared the configured 20 against the API's 10. It is now neither sent nor read back for types that have no preference, so the value is ignored as the docs always claimed.

Also added along the way: `mx_pref` is validated to 0–255, since `setHosts` takes an unsigned byte and 256 was silently truncating to 0 — the most-preferred value, the opposite of the ask.

templates/resources/dns_record.md.tmpl
```diff
@@ -0,0 +1,131 @@
+---
+page_title: "namecheap_dns_record Resource - terraform-provider-namecheap"
+subcategory: "DNS"
+description: |-
+  {{ .Description }}
+---
+
+# namecheap_dns_record (Resource)
```
dm1tr1yvovk: naming is slighly inconsistent - current resource is called
namecheap_domain_records

propose to follow same pattern

templates/resources/dns_record.md.tmpl
```diff
@@ -0,0 +1,131 @@
...
+together on one domain they fight — the zone-level resource will delete records
+this one created (in `OVERWRITE` mode) or fight over ownership (in `MERGE`),
+because neither knows about the other.
+
+- **Use `namecheap_dns_record`** when different records are owned by different
+  people, modules or workspaces, or when you want `for_each` over a map of hosts.
+- **Use `namecheap_domain_records`** when one configuration should own the whole
+  zone, zone-file style — including the guarantee that nothing else exists.
+
+## How a change is applied
```
dm1tr1yvovk: not sure that user should know about internal implementation details
Propose to hide them somewhere in internal docs

docs/resources/domain_host_record.md
dm1tr1yvovk, line 1: why there are two almost identical docs? -> templates/resoruces/dns_record.md.tmpl
maybe we can have only one or write only difference here

namecheap/namecheap_domain_host_record.go
```diff
@@ -0,0 +1,605 @@
...
+				ForceNew:     true,
+				ValidateFunc: validation.StringInSlice(namecheap.AllowedRecordTypeValues, false),
+				Description: fmt.Sprintf("The record type: %s. Changing this forces a new resource. MX and MXE records additionally require the domain's email routing to be set accordingly — see the resource documentation.",
+					strings.Join(namecheap.AllowedRecordTypeValues, ", ")),
+			},
+			"address": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringIsNotEmpty,
+				Description:  "The record's value, whose meaning depends on `type`: an IP address for A/AAAA, a hostname for CNAME/MX/NS, arbitrary text for TXT, a URL for URL/URL301/FRAME. Changed in place.",
```
dm1tr1yvovk: it should not be changes in place
changing type for record should force recreation meaning
delete old and create totally new record

dm1tr1yvovk (changes requested):

dm1tr1yvovk:
One more note, deliberately **not** blocking this PR: there's a read-efficiency follow-up worth doing once this lands.

Because every operation works on the whole zone, a `for_each` fleet of N records costs N identical `getHosts` calls per refresh and ~4 zone-wide calls per record change — linear pressure on the account rate quota as fleets grow. The sibling `terraform-provider-spaceship` provider solves the same shape with a shared per-domain read cache (singleflight reads, invalidated on every write), which cuts a domain refresh from N calls to one; the pattern ports cleanly to SDKv2 via the provider meta.

Filed as #313 with the details so we can manage it separately. Nothing in this PR needs to change for it — the update path's single-read guard (`dnsRecordApplyUpdate`) is already the one-operation version of the same idea.

templates/resources/dns_record.md.tmpl
kurok: Renamed in `8a9b098` — but not to the literal singular, because that name is taken in a way that would hurt.

`namecheap_domain_records` is implemented in `namecheap/namecheap_domain_record.go` as `resourceNamecheapDomainRecords()`. Adding `namecheap_domain_record` would give us two resources, two functions and two doc pages one `s` apart — `resourceNamecheapDomainRecord()` beside `resourceNamecheapDomainRecords()`, `domain_record.md` beside `domain_records.md` — and no file name left for it, since `namecheap_domain_record.go` is the *plural* resource.

So it is **`namecheap_domain_host_record`**: your `namecheap_domain_*` pattern, and the API's own word for these things — `domains.dns.getHosts`/`setHosts`, `DomainsDNSHostRecord` in the SDK.

```hcl
resource "namecheap_domain_host_record" "www" {
  domain   = "example.com"
  hostname = "www"
  type     = "A"
  address  = "203.0.113.10"
}
```

Files and identifiers followed: `namecheap_domain_host_record.go`, `resourceNamecheapDomainHostRecord*`, helpers from the `dnsRecord` prefix to `hostRecord`, `docs/resources/domain_host_record.md`, `examples/resources/domain_host_record/`, plus the README and the importing guide's ID table. Nothing to migrate — the resource has never been in a release.

Happy to switch to plain `namecheap_domain_record` and rename the older file out of the way if you would rather have the exact singular; say so and it is one more commit.

templates/resources/dns_record.md.tmpl
kurok: Agreed — gone from the page in `8a9b098`. The whole "How a change is applied" section (setHosts replacing the zone, the per-domain mutex, the SDK's read-back-and-retry) now lives in the doc comment on `resourceNamecheapDomainHostRecord`, next to the code it describes, where it can also be precise about what the verification does *not* guarantee.

What is left on the page is the part a user has to act on, and nothing about how it works:

> !> **Change a domain from one place at a time.** Namecheap has no per-record API, so a change to any single record rewrites the domain's whole record set. Records managed within one Terraform run are safe — the provider serializes them, and it detects and retries most interference from outside the run — but that window cannot be closed: a second writer landing mid-apply, another `terraform apply` or an edit in the Namecheap dashboard, can still cost you a record silently. Run one apply per domain at a time, and leave the zone alone while it runs.

That warning has to stay in some form: the previous review round found the page claiming a concurrent writer "cannot cost you a record silently", which is false, and silence would repeat the mistake. Six lines instead of seventeen, and no mechanism.

The `mx_pref` note lost its tail for the same reason — it was explaining the post-write check to justify why a value outside `MX` is ignored. It now just says it is ignored.

docs/resources/domain_host_record.md
kurok: They are not two documents — `docs/` is the build output of `templates/`, and both are committed on purpose.

`templates/resources/domain_host_record.md.tmpl` is the source. `tfplugindocs` renders it into `docs/resources/domain_host_record.md`, substituting `{{ .Description }}` from the resource's `Description` in Go and inlining the example files via `{{tffile}}`/`{{codefile}}` — so the examples in the published page are the same files `make examples-validate` type-checks against a real provider binary, rather than HCL pasted into prose that can rot.

The generated tree has to be committed because the Terraform Registry ingests `docs/` at the release tag; it never runs our generator. Every resource here works this way — `domain_records`, `domain_contacts`, `email_forwarding`, `personal_nameserver` all have the same template/output pair, as do the guides.

The two cannot drift: `make docs-check` regenerates and fails the build if the committed `docs/` differs by a byte, or if a generated file is left untracked. It is what caught the missing page when this branch first went up.

So there is nothing to deduplicate — editing `docs/` directly is the thing to avoid, since the next `make docs` overwrites it. I only edit the `.tmpl` and re-run `make docs`.

namecheap/namecheap_domain_host_record.go
kurok: `type` already forces recreation — `ForceNew: true` on it, along with `domain` and `hostname` (lines 63, 70, 80 of this file), and `TestAccMockDomainHostRecordForceNewOnIdentity` pins that a `type` or `hostname` change destroys and recreates rather than editing. The line you commented on is `address`, so I have read this as "`address` should force recreation too" — tell me if you meant something else.

I have not changed it here, and I would rather not, for one concrete reason: replacement is not one API call.

```
ForceNew:  setHosts(zone without the record)   <- the name resolves to nothing here
           setHosts(zone with the new address)
in place:  setHosts(zone with the address swapped)
```

Namecheap has no per-record operation, so a replacement is a full-zone write that removes the record followed by another that adds it back. Between them the name does not resolve at all — an outage on a live `A` record, on a change that is currently atomic from a resolver's point of view. `create_before_destroy` cannot rescue it either: the old and new record would coexist, which is exactly the duplicate this resource refuses to create. In-place is also what every comparable provider does (`aws_route53_record.records`, `cloudflare_record.content`, `azurerm_dns_a_record.records` are all updatable).

That said, your instinct is not wrong, and the tension is real: `address` *is* part of the record's identity everywhere else in this resource — it is in the ID, in the selector, in the import ID — so an in-place change rewrites the resource's own ID, which is unusual. That is worth settling deliberately rather than in a thread, so I have written it up as **#314** with both options and everything that changes if we take `ForceNew` (the ID becomes immutable; the update-path ambiguity guard and the `address` half of the failed-update rollback both become dead code; two tests flip). If you want the outage window instead of the mutable ID, say so on #314 and I will do it there — the switch itself is one line.

kurok:
Review comments addressed in `8a9b098`, replies in the threads. Summary, since one of them touches every file:

**Renamed `namecheap_dns_record` → `namecheap_domain_host_record`.** The literal singular of `namecheap_domain_records` is not usable: that resource lives in `namecheap_domain_record.go` as `resourceNamecheapDomainRecords()`, so `namecheap_domain_record` would leave two resources, two functions and two doc pages one `s` apart. `namecheap_domain_host_record` keeps the `namecheap_domain_*` pattern and borrows the API's own term (`getHosts`/`setHosts`, `DomainsDNSHostRecord`). Files, identifiers (`dnsRecord*` → `hostRecord*`), docs, examples, README and the importing guide followed. Nothing to migrate — never released.

**Moved the zone-rewrite mechanics off the resource page** into the doc comment on the resource. What remains is the one actionable line: change a domain from one place at a time. The `mx_pref` note lost its explanation of the same machinery.

**No change for the two doc trees** — `docs/` is `tfplugindocs` output of `templates/`, committed because the Registry ingests it at the release tag, and kept honest by `make docs-check`. Same pair exists for all five resources.

**No change for `address` being in place**, and reasoning in the thread: replacement here is two full-zone writes with the name unresolvable in between, since Namecheap has no per-record call. The underlying tension is real though — `address` is part of the record's identity everywhere else — so it is filed as #314 to settle on its own, with the one-line switch and its consequences written out. Also noted there: `type` is already `ForceNew`, with a test on it.

Green locally after the rename: `build`, `vet`, `lint`, unit suite, all 34 mock acceptance tests for this resource (68 including subtests across the suite), `docs-check`, `examples-validate` (17 examples).

On the read-caching note above — agreed, and #313 is the right place for it; nothing in this branch changes for it.

templates/resources/dns_record.md.tmpl
dm1tr1yvovk: okay, lets have namecheap_domain_host_record

but you can research other providers and similar resources in cloudflare - they have dns record
https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/dns_record
aws also has dns_record resource

kind of community pattern, but host_record is valid too.

namecheap/namecheap_domain_host_record.go
dm1tr1yvovk: maybe update description to it would be not confusing about change in place

templates/resources/dns_record.md.tmpl
kurok: Thanks — keeping `namecheap_domain_host_record` then. I did go and look, and the research says something worth putting on the record before we settle, because one data point is closer to home than Cloudflare.

**Our own other provider names it exactly the way this PR originally did.** `terraform-provider-spaceship` ships `spaceship_dns_record` (single record) beside `spaceship_dns_records` (the whole custom group) — the same pair, the same split, the same warning about not pointing both at one domain. So `namecheap_dns_record`/`namecheap_domain_records` would have matched the sibling provider, and `domain_host_record` does not.

**The community is split, and not on `dns_record` as such.** What the ecosystem actually agrees on is `<provider>_<scope>_record`, where the scope segment names whatever owns the record:

| Provider | Resource |
|---|---|
| Cloudflare (v5+) | `cloudflare_dns_record` — renamed from `cloudflare_record` |
| AWS | `aws_route53_record` — no `aws_dns_record` exists; the scope is the service |
| Google | `google_dns_record_set` |
| Azure | `azurerm_dns_a_record`, `azurerm_dns_cname_record` — one resource per type |
| DigitalOcean | `digitalocean_record` |
| DNSimple | `dnsimple_zone_record` |
| Linode | `linode_domain_record` |
| OVH | `ovh_domain_zone_record` |

`namecheap_domain_host_record` is in that family — provider, scope, record — and it is what Namecheap itself calls these things. Discoverability is fine either way: the page keeps the `DNS` subcategory and the description opens with "Manages a single DNS host record", so a registry search for "dns record" lands on it.

But the sibling provider is a real argument I would rather you had than not. If cross-provider consistency inside Namecheap matters more than consistency inside this one, the honest name is `namecheap_dns_record` — which is where this branch started, so reverting is one command and no risk. Say the word; otherwise `namecheap_domain_host_record` stands and I will stop poking at it.

namecheap/namecheap_domain_host_record.go
kurok: Done in `2c03e62` — you were right that "Changed in place" was doing too much work in too few words. It read like a claim about the API when the useful thing to say is what the plan does.

The `address` description now says it outright:

> The record's value, whose meaning depends on `type`: an IP address for A/AAAA, a hostname for CNAME/MX/NS, arbitrary text for TXT, a URL for URL/URL301/FRAME. **Changing this does not force a new resource: the existing record is edited, in one write that swaps the value, so the name never stops resolving. The resource id is re-rendered, since the address is part of it.**

`ttl` got the same treatment, and the resource page now names both outcomes in Terraform's own words rather than in prose of mine:

> `domain`, `hostname` and `type` are the record's identity, so changing any of them **destroys the record and creates a new one** — the plan says `-/+ destroy and then create replacement`.
>
> `address`, `ttl` and `mx_pref` are **edited on the record that is already there**: the plan shows `~ update in-place`, one write swaps the value, and the name keeps resolving throughout. Changing `address` also re-renders the resource's `id`, which contains it. If such a change is rejected by the API, the record is left exactly as it was and the next plan proposes the same change again.

The id re-rendering is the part that was genuinely undocumented, and it is the thing your first comment was circling — so #314 stays open for the question of whether that mutability is acceptable at all, or whether `address` should join the identity fields. No behaviour changed here; the `ForceNew` set is untouched.

kurok:
Heads-up on the red `Acceptance test` check on `2c03e62`, since that commit only touched attribute descriptions: the failure is **API throttling on the sandbox account, in a test unrelated to this PR**.

```
--- PASS: TestAccNamecheapDomainHostRecordLifecycle   (66.79s)
--- PASS: TestAccNamecheapDomainHostRecordRoundTrips  (6.25s)
--- PASS: TestAccNamecheapDomainHostRecordMXPreference (132.43s)
--- PASS: TestAccNamecheapDomainHostRecordRefusals    (54.06s)
--- FAIL: TestAccNamecheapEmailForwarding             (293.88s)
    getEmailForwarding failed: after 5 attempts: namecheap: retryable HTTP status
```

Everything this PR adds passed, as did `TestAccNamecheapDomainRecords`, `TestAccDomainValidation` and the data sources. `retryable HTTP status` after five attempts is the sandbox account's hourly request cap, not a flake: the previous live run on `8a9b098` finished at 13:26:48Z and this one started at 13:51Z, so two full live suites landed inside one quota window and the second starved partway through. The same suite passed end-to-end on `8a9b098` 25 minutes earlier, on identical code apart from doc strings.

One side effect worth flagging rather than leaving to be discovered: the email-forwarding test's own cleanup was throttled too (`failed to restore original email forwarding for terraform-provider-test.net`), so that domain may still hold the test's forwarding values until the next successful run of that test resets them.

Re-running the full suite once the quota window has cleared; will report the result here.

dm1tr1yvovk (approved):

dm1tr1yvovk:
Providers acceptance tests returns warnings:
namecheap_domain_host_record_live_test.go:410: MX preference probe on terraform-provider-test.net: changing "mail3.terraform-provider-test.net" from preference 20 to 30 returned concurrent modification detected: DNS zone changed between read and write; the zone then held [pref=20 ttl=1800]

because there would be always in flight modifications with default 10 parallelism

looks like terraform parallelism and there is a need to something with it

kurok:
Correcting my earlier note on the red `Acceptance test`: **the "two suites inside one quota window" explanation was wrong.** I re-ran after a 68-minute idle gap (previous suite ended 14:05:36Z, the re-run started 15:14Z) and it failed at exactly the same place, so spacing is not the variable.

What the failure *is*, precisely — the error text hides it:

`namecheap: retryable HTTP status` is the SDK's sentinel for **HTTP 405, which Namecheap returns to signal rate limiting** ([`transport.go:34-38`](https://github.com/namecheap/go-namecheap-sdk/blob/main/namecheap/transport.go)), and it is raised only after all five attempts with backoff came back 405. So it is rate limiting, but a cumulative one — the documented caps are 20/min, 700/hour and 8000/day, and today's repeated full live suites eat the longer-window budgets. The suite runs `TestAccNamecheapEmailForwarding` last, roughly 11 minutes in, so it is the test that finds the budget already spent, whichever earlier tests spent it.

Two runs, same shape, everything else green:

```
15:15:25 --- PASS: TestAccNamecheapDomainHostRecordLifecycle    (58.93s)
15:15:40 --- PASS: TestAccNamecheapDomainHostRecordRoundTrips   (15.39s)
15:17:52 --- PASS: TestAccNamecheapDomainHostRecordMXPreference (131.78s)
15:18:46 --- PASS: TestAccNamecheapDomainHostRecordRefusals     (54.02s)
15:23:04 --- PASS: TestAccNamecheapDomainRecords                (257.87s, 10 subtests)
15:23:24 --- PASS: TestAccDomainValidation                      (20.00s)
15:29:05 --- FAIL: TestAccNamecheapEmailForwarding              (341.05s)
```

That is still not this PR: the failing test touches none of this code, everything the PR adds passes live, and the commit differs from a fully green run only in attribute descriptions.

**Not re-running again today** — another attempt would fail the same way and spend more of the budget. Worth doing instead, separately from this PR:

- Whoever owns the sandbox credentials could check the account's daily consumption; if the 8000/day cap is the one being hit, no amount of waiting inside today helps.
- `terraform-provider-test.net` has now had its email forwarding left unrestored twice (the cleanup 405s too, `namecheap_email_forwarding_live_test.go:172`), so that domain needs a manual reset or a successful run of that test.
- The terminal message deserves to say "rate limited (HTTP 405) after 5 attempts" rather than "retryable HTTP status" — it cost me a wrong diagnosis, and it will cost the next person the same. SDK-side, not here.

PR #315, opened 2026-08-05
ci: gate CI jobs on changed paths to skip redundant acceptance runs

dm1tr1yvovk:
## Why

A single code change currently runs the full pipeline **4 times** on its way to a release: on the PR, on its merge to `master`, on the release-please Release PR, and on that PR's merge — although the Release PR only touches `CHANGELOG.md` and `.release-please-manifest.json`. Two of the four runs re-test code that was already tested (tracked in DEVOPS-21842). In this repository that waste is not just hosted-runner minutes: each redundant run also performs a full start/stop cycle of the **self-hosted EC2 sandbox runner** (`t3.medium` + the whitelisted Elastic IP) and a ~10-minute live sandbox suite, so every gated-away run saves real AWS cost and frees a slot in the single-EIP acceptance queue.

Same pattern as namecheap/terraform-provider-jenkins#200.

## What

- Adds a `changes` gate job (`dorny/paths-filter@v4.0.2`, pinned by SHA per `SECURITY_COMPLIANCE.md`, `predicate-quantifier: every`). All test jobs — `check`, `docs`, both security scans, the security summary, the `acceptance_mock` matrix, and the EC2 `start-runner`/`acceptance_test` pair — take `needs: changes` + an `if:` on its `code` output, combined with their pre-existing fork/Dependabot/event conditions. They skip when the diff touches only root-level markdown (`*.md`), `LICENSE`, or `.release-please-manifest.json`.
- `docs/**`, `templates/**` and `examples/**` are treated as code and run the full suite: `docs/` is rendered by the Terraform Registry from the repo at each release tag, so it is a user-visible deliverable, and every example is type-checked by the `docs` job.
- **`stop-runner` is intentionally NOT gated**: its existing `always() && needs.start-runner.outputs.ec2-instance-id != ''` condition already skips cleanly when `start-runner` was skipped by the gate (the output is empty), and still stops any runner that was actually started — no change to the EC2 lifecycle semantics, and the nightly `cleanup-ec2-runners.yml` backstop is unaffected.
- **`workflow_dispatch` bypasses the filter** (always `code=true`): manual dispatch is a maintainer explicitly asking for a run — it is the documented path for live acceptance on Dependabot branches (`gh workflow run CI --ref <branch>`), and paths-filter has no base to diff against on that event anyway.
- Adds an always-run **`CI OK`** summary job (`needs:` every job, fails on `failure`/`cancelled`, passes on `success`/`skipped`) so a fully-skipped run still concludes `success` — required by `versioning.yml`'s `workflow_run` gate — and can serve as the single required check.
- Adds `CI.md` (human-oriented pipeline map + editing invariants) and updates `CONTRIBUTING.md`, `README.md`, `RELEASE.md`, `SECURITY_COMPLIANCE.md`, and `CLAUDE.md` where they describe when CI jobs run.

`.release-please-config.json` was checked before putting the manifest in the ignore list: `release-type: go` with no `extra-files`, so the Release PR touches only `CHANGELOG.md` + `.release-please-manifest.json`. If `extra-files` is ever added, the manifest must come out of the ignore list (noted in `CI.md`'s invariants).

## Why not `paths-ignore` on the triggers

1. The `Check` / `Security scan (Trivy)` / `Security scan (govulncheck)` checks are required by the "Protect master" ruleset, and a path-filtered required check is never reported on PRs touching only excluded files — such a PR could never merge. Checks skipped via `if:` **do** satisfy branch protection.
2. If CI didn't trigger at all on the Release-PR merge push, `versioning.yml`'s `workflow_run: [CI]` would never fire and release-please would never create the tag, breaking the release-please → tag → GoReleaser chain.

## Branch protection

The "Protect master" ruleset currently requires: `Check`, `Security scan (Trivy)`, `Security scan (govulncheck)` (all from this workflow), and `security/snyk (Namecheap)`. **After this PR merges, "CI OK" must be added to that list.** This is mandatory, not optional: the gate fails open — if the `changes` job itself fails (API flake, runner loss), every test job reports "skipped", and skipped checks satisfy the existing required checks — so a failing "CI OK" is the only thing that blocks such a PR from merging untested. The existing required checks can stay (they'll report "skipped" on docs-only PRs, which passes) or be replaced by "CI OK" alone.

## Verification

- `actionlint` (v1.7.10) passes on the modified workflow.
- This PR itself exercises the `code=true` path (workflow + markdown changes → full suite, EC2 acceptance included).
- The next release-please Release PR will exercise the skip path: expect every test job "skipped", "CI OK" green, and a normal tag + GoReleaser run after its merge.
- Dependabot PRs are unchanged: `check`/`docs`/scans + `acceptance_mock` run (their diffs touch `go.mod`/workflows — code), the EC2 jobs stay skipped, and the manual `workflow_dispatch` path now explicitly bypasses the gate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,14 @@
 name: CI
 
+# No workflow-level paths-ignore here, on purpose. This workflow's jobs are
+# required status checks, and a path-filtered required check is never reported
+# on PRs that touch only excluded files, which blocks them from merging
+# permanently. It would also stop versioning.yml's `workflow_run: [CI]` trigger
+# from firing on pushes that only touch release files, breaking the
+# release-please -> tag -> GoReleaser chain. Instead, the `changes` job below
+# detects what changed and the expensive jobs skip themselves via `if:` —
+# skipped checks DO satisfy branch protection, and the workflow still completes
+# so workflow_run still fires.
 on:
   # Post-merge run only: PR branches are covered by pull_request, so each PR
   # runs CI once. Pre-PR branches: use workflow_dispatch.
@@ -36,9 +45,54 @@ env:
 # queue step in start-runner below.
 
 jobs:
+  # Gate job: decides whether the jobs below actually need to run for this
+  # change. `code` is false only when the diff touches nothing but root-level
+  # markdown (README/CHANGELOG/...), LICENSE, or the release-please manifest —
+  # most notably the release-please Release PR (CHANGELOG.md +
+  # .release-please-manifest.json) and its merge to master, which previously
+  # re-ran the full pipeline — self-hosted EC2 acceptance run included — twice
+  # per release on code that was already tested. docs/** and templates/**
+  # deliberately count as code: docs/ is published to the Terraform Registry at
+  # each release tag, so doc changes get the full pipeline like any
+  # user-visible change.
+  changes:
+    name: Detect Changes
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read # paths-filter reads PR files via the API on pull_request events
+    outputs:
+      # workflow_dispatch is a maintainer explicitly asking for a full run (the
+      # documented path for live acceptance on Dependabot branches — see
+      # CLAUDE.md), and paths-filter has no base to diff against there anyway,
+      # so dispatch runs always count as code and bypass the filter entirely.
+      code: ${{ github.event_name == 'workflow_dispatch' && 'true' || steps.filter-code.outputs.code }}
+    steps:
+      # Checkout is required for push events, where paths-filter diffs with git
+      # (pull_request events use the API and would not need it).
+      - uses: actions/checkout@v7
+        if: ${{ github.event_name != 'workflow_dispatch' }}
+
+      # With predicate-quantifier=every a file counts as `code` only if it
+      # matches ALL patterns, i.e. it is none of the ignored kinds. `*.md`
+      # (no globstar) matches root-level markdown only, so nested markdown —
+      # docs/**, templates/** — still counts as code on purpose.
+      - uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4.0.2
+        if: ${{ github.event_name != 'workflow_dispatch' }}
+        id: filter-code
+        with:
+          predicate-quantifier: 'every'
+          filters: |
+            code:
+              - '!*.md'
+              - '!LICENSE'
+              - '!.release-please-manifest.json'
+
   check:
     name: Check
     runs-on: ubuntu-latest
+    needs: changes
+    if: ${{ needs.changes.outputs.code == 'true' }}
     steps:
       - uses: actions/checkout@v7
 
@@ -125,6 +179,8 @@ jobs:
     # dev_overrides config, so this job is fork-safe.
     name: Docs
     runs-on: ubuntu-latest
+    needs: changes
+    if: ${{ needs.changes.outputs.code == 'true' }}
     steps:
       - uses: actions/checkout@v7
 
@@ -163,6 +219,8 @@ jobs:
     # audits) can consume a signed inventory without re-running scans.
     name: Security scan (Trivy)
     runs-on: ubuntu-latest
+    needs: changes
+    if: ${{ needs.changes.outputs.code == 'true' }}
     steps:
       - uses: actions/checkout@v7
 
@@ -252,6 +310,8 @@ jobs:
     # so detection stays current regardless of the pinned binary version.
     name: Security scan (govulncheck)
     runs-on: ubuntu-latest
+    needs: changes
+    if: ${{ needs.changes.outputs.code == 'true' }}
     steps:
       - uses: actions/checkout@v7
 
@@ -307,9 +367,11 @@ jobs:
     # in the summary via needs.*.result, so a green comment can never mask a red
     # gate.
     name: Security scan summary
-    needs: [security, govulncheck]
+    needs: [changes, security, govulncheck]
     # Run even when a gate above failed — that's when the summary matters most.
-    if: ${{ !cancelled() }}
+    # But not when the changes gate skipped the scans entirely: there are no
+    # scan artifacts to summarize on a docs-only change.
+    if: ${{ !cancelled() && needs.changes.outputs.code == 'true' }}
     runs-on: ubuntu-latest
     permissions:
       contents: read # checkout + read go.mod for the dependency overview
@@ -413,8 +475,8 @@ jobs:
     # Dependabot PRs (same). Same-repo, non-Dependabot PRs run the live sandbox
     # suite instead, so those two never both run on one PR.
     name: Mock acceptance (${{ matrix.engine }} ${{ matrix.version }})
-    needs: [check]
-    if: ${{ github.event_name == 'pull_request' && (github.event.pull_request.head.repo.full_name != github.repository || github.actor == 'dependabot[bot]') }}
+    needs: [changes, check]
+    if: ${{ needs.changes.outputs.code == 'true' && github.event_name == 'pull_request' && (github.event.pull_request.head.repo.full_name != github.repository || github.actor == 'dependabot[bot]') }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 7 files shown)


dm1tr1yvovk:
Note on the red "Acceptance test" / "CI OK": the live sandbox suite is failing today with `after 5 attempts: namecheap: retryable HTTP status` (API throttling) — the same error failed the latest `master` run ([30999878228](https://github.com/namecheap/terraform-provider-namecheap/actions/runs/30999878228), the #310 merge, which predates this PR), so it is unrelated to this change. Two runs of this PR confirmed the gate wiring works on the code path: Detect Changes → full pipeline, EC2 runner started and stopped cleanly both times, mock matrix skipped as before, and "CI OK" correctly aggregated the acceptance failure. Will re-run once the sandbox recovers. Reminder: acceptance flakes here need a full re-run, not "re-run failed jobs" — stop-runner has already stopped the runner the failed job's label points at.

dm1tr1yvovk:
**Production validation from the identical gate in terraform-provider-spaceship** (merged as [spaceship#130](https://github.com/namecheap/terraform-provider-spaceship/pull/130)): v1.0.0 was released end-to-end through the new pipeline.

- Skip path on master pushes — [empty Release-As commit: 26s, all test jobs skipped, CI OK success](https://github.com/namecheap/terraform-provider-spaceship/actions/runs/31010691414), [Release PR merge: ~30s](https://github.com/namecheap/terraform-provider-spaceship/actions/runs/31011003106), [markdown-only merge: ~30s](https://github.com/namecheap/terraform-provider-spaceship/actions/runs/31011542487) — vs ~5 min full runs before. On this repo the same skips also avoid spinning up the EC2 acceptance runner for release-please traffic.
- Release chain intact: the skip runs' `success` conclusions triggered Versioning normally → tag `v1.0.0` → [GoReleaser](https://github.com/namecheap/terraform-provider-spaceship/actions/runs/31011046451) published the [signed release with 16 assets](https://github.com/namecheap/terraform-provider-spaceship/releases/tag/v1.0.0).
- Full-suite path confirmed for code and `docs/` changes ([docs-prep merge ran everything](https://github.com/namecheap/terraform-provider-spaceship/actions/runs/31009098507)).

Remaining here: re-run the workflow once the sandbox throttling clears (whole run, not failed-jobs-only — the runner label dies after stop-runner), and add "CI OK" to the master ruleset's required checks after merge. Tracked in DEVOPS-21842.

.github/workflows/ci.yml
```diff
@@ -878,3 +949,38 @@ jobs:
...
+  #    versioning.yml via workflow_run) ambiguous. This job always executes,
+  #    so the run always concludes success or failure.
+  # 2. Branch protection can require just this one check instead of listing
+  #    every job: it fails if any needed job failed or was cancelled, and
+  #    passes when jobs succeeded or were legitimately skipped — by the
+  #    changes gate, or by the pre-existing fork/Dependabot/event conditions
+  #    on acceptance_mock and the EC2 jobs.
+  # Every job in this workflow MUST be listed in `needs` below; a job missing
+  # from the list is invisible to the required check.
+  ci-ok:
```
kurok, line 970: **Merge-order hazard — I'd add "CI OK" to the ruleset *before* merging this, not after.**

The description says "After this PR merges, 'CI OK' must be added to that list." I checked the current state of the "Protect master" ruleset (id `14733758`) and it requires:

```
Check
Security scan (govulncheck)
Security scan (Trivy)
security/snyk (Namecheap)
```

No "CI OK". Which means the window between this merging and someone editing the ruleset is exactly the fail-open case you documented: if the `changes` job errors (API flake, runner loss), `check` / `security` / `govulncheck` all report `skipped`, skipped satisfies those three required checks, Snyk is independent and stays green — and an untested PR is mergeable. Today that can't happen, because `check` failing to run at all leaves its check unreported.

So this PR briefly *weakens* the gate it exists to strengthen, and the mitigation is a manual step that has to happen in the right order. For comparison, on terraform-provider-jenkins#200 "CI OK" is already in the required list, so that one merges into a safe state.

Adding the context to the ruleset before merge is harmless — "CI OK" simply won't be reported until the workflow exists, and you'd re-run once. Alternatively hold the merge behind whoever owns the ruleset.

.github/workflows/ci.yml
```diff
@@ -878,3 +949,38 @@ jobs:
...
+      - check
+      - docs
+      - security
+      - govulncheck
+      - security-report
+      - acceptance_mock
+      - start-runner
+      - acceptance_test
+      - stop-runner
+    if: always()
```
kurok: Same note as on terraform-provider-jenkins#200: `!cancelled()` reads better here than `always()`. An explicit `if:` that isn't `success()`-based still runs when a dependency **failed**, so the blocking behaviour is unchanged, and `result == 'cancelled'` on an individually-cancelled job is still caught. The difference is only when the whole run is cancelled — today that spins up a runner for this job and posts a red "CI OK" on the abandoned SHA.

Worth stating explicitly for this repo: this does **not** touch the EC2 teardown. `stop-runner` keeps its own `always()` and its own `needs`, so a cancelled run still tears the instance down exactly as it does now.

```suggestion
    if: ${{ !cancelled() }}
```

kurok (commented):
Checked the parts the design leans on and they hold up:

- **The premise is real.** `.release-please-config.json` is `release-type: go` with no `extra-files`, and the last three Release PRs (#312, #309, #304) touch exactly `.release-please-manifest.json` + `CHANGELOG.md` — both in the ignore list. So the skip path will fire.
- **`stop-runner` reasoning is correct.** When `start-runner` is skipped, `needs.start-runner.outputs.ec2-instance-id` is empty, so `always() && ... != ''` skips cleanly. Leaving it ungated is right, and the EC2 lifecycle is genuinely untouched.
- **The `workflow_dispatch` bypass expression is correct.** `A && 'true' || B` returns `'true'` on dispatch and falls through to the filter output otherwise, including the `'false'` string case.
- **`ci-ok`'s `needs` list is complete** — all 10 non-`ci-ok` jobs are listed.
- **The queue isn't affected.** `action-workflow-queue` lives inside `start-runner`, so a gated-away run never takes the lock; docs-only runs now clear the EIP queue faster than before.

One blocking-ish thing about merge order, plus the same two `ci-ok` nits I left on namecheap/terraform-provider-jenkins#200.

.github/workflows/ci.yml
```diff
@@ -878,3 +949,38 @@ jobs:
...
+    steps:
+      - name: Check results of all jobs
+        env:
+          NEEDS: ${{ toJSON(needs) }}
+        run: |
+          echo "$NEEDS" | jq -r 'to_entries[] | "\(.key): \(.value.result)"'
+          echo "$NEEDS" | jq -e 'to_entries | all(.value.result == "success" or .value.result == "skipped")' > /dev/null
```
kurok: Two things, same as terraform-provider-jenkins#200 but they bite harder here.

**The `needs` list is hand-maintained.** The comment above says "Every job in this workflow MUST be listed in `needs` below; a job missing from the list is invisible to the required check" — which is exactly right, and exactly the kind of rule that decays. This file is 986 lines with 11 jobs and a fair amount of conditional machinery; a job added near `acceptance_mock` is a long way from this list. And the failure mode is silent in the worst direction: "CI OK" green while the unwatched job is red.

**And `jq -e ... > /dev/null` exits 1 with no message.** The table on the line above is in the log, but an annotation puts the reason on the checks UI.

```suggestion
    steps:
      - uses: actions/checkout@v7

      # "CI OK" only guards the jobs listed in its `needs`. A job added to this
      # workflow but not added there would leave this check green while it
      # fails — the exact gap this job exists to close — so assert the two
      # lists match rather than relying on the comment above surviving review.
      - name: Assert every job is watched by this gate
        run: |
          watched=$(yq -r '.jobs.ci-ok.needs[]' .github/workflows/ci.yml | sort)
          all=$(yq -r '.jobs | keys | .[]' .github/workflows/ci.yml | grep -vx 'ci-ok' | sort)
          if ! diff <(echo "$all") <(echo "$watched"); then
            echo "::error::ci.yml jobs and ci-ok's needs list have drifted (< missing from needs, > stale entry)."
            exit 1
          fi

      - name: Check results of all jobs
        env:
          NEEDS: ${{ toJSON(needs) }}
        run: |
          echo "$NEEDS" | jq -r 'to_entries[] | "\(.key): \(.value.result)"'
          if ! echo "$NEEDS" | jq -e 'to_entries | all(.value.result == "success" or .value.result == "skipped")' > /dev/null; then
            echo "::error::One or more CI jobs failed or were cancelled — see the job results above."
            exit 1
          fi
```

Caveats, since I couldn't run this: it adds a checkout to `ci-ok` (a couple of seconds — and `ci-ok` already inherits `contents: read`), and it assumes `yq` is present on the `ubuntu-latest` image. I believe it is but I did not verify it, so worth confirming on the first run — otherwise a `python3 -c` one-liner does the same job.

PR #327, opened 2026-08-26
ci: drop ec2 runner warm-pool reuse in favor of terminate-per-run

dm1tr1yvovk:
## Why

DEVOPS-22119 investigated the v4 `ec2-github-runner` reliability and speed regressions in this repo's acceptance CI. Measurements across 12 v4-era runs vs 7 v3-era runs showed the warm pool (`reuse: stop`, adopted with #279) delivers **negative** net value here:

| Metric | v3 (terminate) | v4 warm pool |
|---|---|---|
| `Start EC2 runner` step | 59-60 s | **70 s — warm hit and cold miss identical** |
| stop-runner job | ~6 s (fire-and-forget terminate) | 22-152 s (avg ~60 s, blocks on `stopped`) |
| Net per run | baseline | **≈ +65 s** |

The premise behind the warm pool (2-4 min cold-boot tax) does not hold on our baked `nc-al2023` AMI: a cold launch registers in ~70 s, and the ~10 s a warm boot actually saves is smaller than the action's fixed overheads (unconditional 30 s bootstrap wait, 15 s instance-running waiter floor, 3×10 s warm stability confirmations). Parking the instance also introduced the EIP-drift failure class that required the consumer-side "Pre-attach sandbox EIP" workaround step.

## What

- **ci.yml**: drop `reuse: stop` / `reuse-pool-tag` from `start-runner` and `stop-runner` (back to the action's default terminate-per-run); delete the "Pre-attach sandbox EIP" step — it only existed because a warm restart skipped the action's own `eip-allocation-id` association, which now runs on every launch.
- **cleanup-ec2-runners.yml**: remove the nightly full-drain pass (its only purpose was emptying the warm pool); keep the 30-minute leak reaper for crashed/cancelled leftovers. `workflow_dispatch` keeps the `dry_run` preview.
- **CI.md / CONTRIBUTING.md / SECURITY_COMPLIANCE.md / scripts/hygiene-sweep.sh**: update the runner-lifecycle description. Hygiene sweeps are kept as defense in depth. `ec2:StopInstances` / `ec2:StartInstances` / `ec2:ModifyInstanceAttribute` IAM grants are no longer required and can be revoked.

Every run now starts from a pristine disk, which also removes the job-N+1-runs-on-job-N's-disk consideration the warm pool had to document and sweep around.

## Verification

- Both workflows parse as valid YAML.
- `ci-ok`'s `needs:` list still matches the job list exactly (the gate's own assertion passes: 10 jobs watched).
- No `reuse`/warm-pool references remain in workflows or scripts; remaining doc mentions are deliberate historical context for the removal.
- The acceptance pipeline on this PR exercises the changed `start-runner`/`stop-runner` path end to end (cold launch + terminate).

Refs: DEVOPS-22119, #279 (warm-pool introduction).

## EIP handover race (second commit)

With terminate-per-run, the EIP is auto-released only when the old instance reaches `terminated`, while the workflow queue only waits for the older *run* to finish — so a back-to-back run could reach the launch while the previous instance is still shutting down with the EIP attached. The action's `AssociateAddress` does not set `AllowReassociation` and downgrades the failure to a warning (`src/aws.js:920-927`), which would surface only later at the "Verify sandbox EIP" gate. The second commit adds a **"Wait for the sandbox EIP to be free"** step in `start-runner`: a single read-only `DescribeAddresses` call when the EIP is already free (the common case), polling up to 120s only under genuine back-to-back contention, warning-then-proceed on timeout. No new IAM required.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -585,6 +585,33 @@ jobs:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: ${{ secrets.AWS_REGION }}
+      - name: Wait for the sandbox EIP to be free
+        # stop-runner's terminate is fire-and-forget, and the workflow queue
+        # above only waits for the older RUN to finish -- not for its
+        # instance to reach `terminated`, which is when AWS auto-releases
+        # the EIP. A back-to-back run can therefore arrive here while the
+        # previous instance is still shutting down with the EIP attached.
+        # The action's own AssociateAddress does not set AllowReassociation
+        # and swallows the failure with a warning, so the run would only
+        # fail later (and less obviously) at acceptance_test's "Verify
+        # sandbox EIP" gate. Poll here instead: a single read-only
+        # DescribeAddresses call when the EIP is already free (the common
+        # case), a short wait only under genuine back-to-back contention.
+        # On timeout, proceed anyway -- the action will warn and the verify
+        # gate still fails cleanly.
+        run: |
+          for i in $(seq 1 24); do
+            assoc="$(aws ec2 describe-addresses \
+              --allocation-ids "$SANDBOX_EIP_ALLOCATION_ID" \
+              --query 'Addresses[0].AssociationId' --output text)"
+            if [ -z "$assoc" ] || [ "$assoc" = "None" ]; then
+              echo "Sandbox EIP is free."
+              exit 0
+            fi
+            echo "Sandbox EIP still associated (${assoc}); waiting for the previous instance to finish terminating (${i}/24)..."
+            sleep 5
+          done
+          echo "::warning::Sandbox EIP still associated after 120s; proceeding — the launch will warn and the acceptance \"Verify sandbox EIP\" gate will fail cleanly if the association cannot be made."
       - name: Start EC2 runner
         id: start-ec2-runner
         uses: namecheap/ec2-github-runner@54ee3e45b519c062993c2b01eb77552fcc329b9d # v4.0.2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -561,14 +561,13 @@ jobs:
       public-ip: ${{ steps.eip-ip.outputs.public-ip }}
     steps:
       - name: Wait for older acceptance runs to finish (single sandbox EIP)
-        # The acceptance pipeline uses one whitelisted Elastic IP and one
-        # warm-pool instance; two runs at once would both try to attach that
-        # EIP and start that instance, fighting over them. Queue this run (FIFO,
-        # no cancellation) behind any older in-progress CI run so only one holds
-        # the sandbox EIP at a time. The older run frees it when its stop-runner
-        # stops the pooled instance (reuse: stop, not terminate — the EIP stays
-        # attached to the stopped instance); this run's "Pre-attach sandbox EIP"
-        # step then re-attaches it before start-up.
+        # The acceptance pipeline uses one whitelisted Elastic IP; two runs at
+        # once would both try to associate it. Queue this run (FIFO, no
+        # cancellation) behind any older in-progress CI run so only one holds
+        # the sandbox EIP at a time. The older run frees it when its
+        # stop-runner terminates the instance (termination auto-disassociates
+        # the EIP back to the account); this run's launch then associates it
+        # via eip-allocation-id.
         uses: ahmadnassri/action-workflow-queue@542658b3a8270cac81ae15d401b0d974732808ac # v1.2.0
         with:
           timeout: 1800000 # 30 min cap for a backlog of queued acceptance runs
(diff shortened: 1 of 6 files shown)


kurok:
Why you decide to remove warm pool feature - when a lot of changes happens this is very slow down work of CI
if you need 3-4 PR in row to make it will be very slow

dm1tr1yvovk:
## Where the time went in this PR's CI runs

Two full acceptance passes ran back-to-back (one per commit), serialized on the single sandbox EIP — which makes this PR an accidental live demo of both the fix and the remaining bottleneck.

### Run 1 — commit `1269233` ([32972997452](https://github.com/namecheap/terraform-provider-namecheap/actions/runs/32972997452)), 15m48s total

| Phase | Duration | Notes |
|---|---|---|
| Hosted checks gating the EC2 leg (`check` et al.) | 1m47s | `start-runner` needs `check`/`security`/`govulncheck` |
| Start self-hosted EC2 runner | 1m19s | cold launch (no queue — nothing ahead of it) |
| Acceptance test | 12m17s | **12m05s of that is `make testacc` against the live sandbox API** — every other step (checkout, sweeps, toolchain, EIP verify) is seconds |
| Stop self-hosted EC2 runner | **8s** | fire-and-forget terminate; was 22–152s (avg ~60s) with the warm pool's wait-for-`stopped` |

### Run 2 — commit `d6a637d` ([32973537105](https://github.com/namecheap/terraform-provider-namecheap/actions/runs/32973537105)), start-runner 9m52s

| Step | Duration | Notes |
|---|---|---|
| Wait for older acceptance runs (workflow queue) | **8m37s** | FIFO wait for run 1 to finish (13:21:31 → 13:30:08) |
| Wait for the sandbox EIP to be free (new guard) | **1s** | `Sandbox EIP is free.` on the first poll — run 1's instance had terminated 24s earlier (13:29:45), so release had already landed |
| Start EC2 runner (action step) | 60s | `run_instance` 1s → instance-running wait → fixed 30s bootstrap wait → `wait_for_runner outcome:online elapsed_s:10, confirm_checks:2` |

### Takeaways

1. **The runner lifecycle is no longer where the time goes.** Cold launch 60–79s, stop 8s, EIP guard 1s. The dominant cost per run is the ~12-minute live suite itself (sandbox API latency × test count) — unaffected by this PR either way.
2. **The single biggest avoidable cost was the 8m37s queue wait behind a superseded run.** Run 1 was obsolete the moment `d6a637d` was pushed, yet it ran to completion and run 2 waited for it. This is exactly the DEVOPS-22119 item-5 case for per-ref `concurrency` with `cancel-in-progress` on PR events (keeping the FIFO queue for cross-ref EIP serialization) — would have saved ~9 minutes here. Proposed as a follow-up PR to keep this one single-purpose.
3. **The EIP handover race didn't trigger** (24s gap between the old instance reaching `terminated` and the new run's launch was enough), and the guard confirmed the free state in one read-only call — it costs nothing in the common case and only waits under genuinely tight back-to-back timing.

dm1tr1yvovk:
@kurok this feature causes more harm when gives benefits.
there are a lot of bugs and I need to cancel one build multiple times just to schedule agent
allocation could take hours.
go-namecheap-sdk does not use this feature and does not have any such problems with same version of runner

dm1tr1yvovk:
@kurok second thing
even if the agent is allocated very fast, we still have api limits
so even having fast agent, we would not get fast build because of retries or api failures

kurok:
checking

dm1tr1yvovk:
check the latest run - https://github.com/namecheap/terraform-provider-namecheap/actions/runs/32973537105/job/98201843671?pr=327

it is not fast

kurok:
One thing I think is a real blocker for the cutover, plus the reaper knob that goes with it.

**A stopped warm-pool instance still holding the EIP after merge → acceptance CI red for up to ~24h.**

master (and every un-rebased branch) still runs the old `reuse: stop` pipeline, so any acceptance run that lands between the last nightly drain and this merge parks a *stopped* `sandbox-acceptance` instance with `eipalloc-1796f61b` still attached — that's the normal end state of the old model (the pre-attach step put it there with `--allow-reassociation`, stop-runner only did `StopInstances`, and an EIP stays associated with a stopped instance). After this PR:

- the surviving leak reaper runs `mode: cleanup` with the action's default `reaper-stopped-max-age` (1440 min). In `cleanup.js` `decideReap` the `stopped` branch returns `stopped-within-max-age` *before* the `runner-not-registered` check ever runs, so that instance is skipped for ~24h (age counts from the `started-at` tag the last warm start rewrote);
- the new "Wait for the sandbox EIP to be free" step polls 120s, warns, proceeds → the action's `AssociateAddress` (no `AllowReassociation`) fails with a warning → the run dies at "Verify sandbox EIP". Every run, until that clock expires;
- the nightly-drain cron and the `mode: nightly-drain` dispatch are deleted in the same diff, so there's no in-repo lever left to drain it (`workflow_dispatch` of the new file also runs at 1440).

Right now the pool looks empty (both runs on this PR got `Sandbox EIP is free` on the first poll), so whether this bites depends purely on merge ordering — another PR merging to master first, an in-flight run at merge time, or a re-run of a pre-merge run is enough.

Fix is one line: add `reaper-stopped-max-age: 1` to the Leak reaper step. In terminate-per-run mode nothing legitimate is ever `stopped` (cold start / terminate / TTL / cleanup never call `StopInstances`), and the legacy instance carries the `managed` + `repository` tags the reaper filters on (there's no pool-tag filter), so it gets terminated on the next 30-min tick and the EIP is released. Draining via the old workflow's `nightly-drain` dispatch right before merging works too, but the reaper input makes it self-healing.

While you're on those inputs: `max-age-minutes` is also at the default 120, and with a single EIP a leaked *running* instance (start-runner killed after launch, wedged job) isn't a billing leak anymore, it's an outage for every acceptance run for up to ~2.5h (120 + 30-min cron + schedule lag). `max-age-minutes: 20` is safe — busy runners are never reaped and the 15-min grace floor protects in-flight starts. None of these numbers made it into the rewritten docs (they all say "default thresholds" / "every 30 minutes"), so worth stating the actual worst-case window there.

kurok (changes requested):
check my comment for major issues and all medium i made iseparate ssues

dm1tr1yvovk:
Verified against the action source and confirmed on all points: `decideReap`'s `stopped` branch (`src/cleanup.js:26-31`) returns `stopped-within-max-age` before the runner check, age counts from the `started-at` tag the last warm start rewrote, and the launch-side `AssociateAddress` has no `AllowReassociation` and downgrades failure to a warning — so a legacy parked instance would indeed redline every acceptance run at the "Verify sandbox EIP" gate for up to ~24h with no in-repo drain lever left.

Fixed in 2aae8af:

- `reaper-stopped-max-age: 1` on the leak reaper — in terminate-per-run mode nothing legitimate is ever `stopped`, so any stopped managed instance in this repo's scope is a legacy leftover or an anomaly; drained on the next 30-min tick, which releases the EIP. Self-healing regardless of merge ordering.
- `max-age-minutes: 20` — a leaked *running* instance now blocks acceptance for at most ~1h (20 min + cron cadence + schedule lag) instead of ~2.5h. Relied on the same properties you cited: busy runners are never reaped, 15-min grace floor covers in-flight starts, and in this pipeline the acceptance job picks the runner up within seconds of registration, so 20 idle minutes is genuinely orphaned.
- Actual worst-case windows are now stated in the workflow header and the SECURITY_COMPLIANCE.md leak-reaper bullet instead of "default thresholds", and the dry-run placeholder echoes the real values.

#328–#335 stay as separate follow-ups per your split.

kurok (approved):

